PR #710, opened 2026-01-13
fix(vite-plugin): add smrtDependencies discovery to OXC scanner path

### LINKED ISSUE #690: db:migrate doesn't detect STI subclass columns for schema migration
## Summary

When using Single Table Inheritance (STI), the `smrt db:migrate` command doesn't detect columns that are defined on subclasses. This causes schema drift when subclass properties are serialized to the shared table but the columns don't exist.

## Problem

STI subclasses (e.g., `Meeting` extends `Event`, `WeatherForecast` extends `Event`) have additional properties that get serialized to the parent table (`events`). However, `db:migrate` only knows about columns defined in the base class manifest.

### Example

The `Event` manifest defines columns like `name`, `start_date`, `end_date`, etc.

But when a `WeatherForecast` is serialized, it writes additional columns:
- `temperature_high`
- `temperature_low`
- `feels_like`
- `humidity`
- `pressure`
- `wind_speed`
- etc.

Similarly, `Meeting` writes:
- `council_id`
- `agenda_url`
- `minutes_url`
- etc.

These columns are not detected by `db:migrate` because they're not in the base `Event` manifest.

## Current Behavior

```
$ smrt db:migrate --dry-run

⚠️  Type mismatches detected (require manual intervention):
   events.metadata: expected TEXT, found jsonb
   ...

🔨 Applying 6 migration(s)...
  ✗ add_column failed: Failed to execute raw query
```

The migrate command:
1. Doesn't detect missing STI subclass columns
2. Only sees columns from the base class manifest
3. Results in runtime errors: `column "temperature_high" of relation "events" does not exist`

## Expected Behavior

`db:migrate` should:
1. Discover all STI siblings for each table (it already does this via `discoverSTISiblingsSync`)
2. Merge fields from all STI subclasses into the parent table's expected schema
3. Detect missing columns from ANY subclass, not just the base class
4. Generate appropriate `ALTER TABLE ADD COLUMN` statements

## Workaround

Currently requires manual intervention:
```sql
ALTER TABLE events ADD COLUMN IF NOT EXISTS temperature_high TEXT;
ALTER TABLE events ADD COLUMN IF NOT EXISTS humidity TEXT;
-- ... many more columns
```

## Environment

- smrt-cli version: latest
- Database: PostgreSQL
- STI tables affected: `events`, `contents`, `profiles`, `places`, `agents`

## Related

The manifest loader already discovers STI siblings:
```
[manifest-loader] discoverSTISiblingsSync called for collection: events
[manifest-loader] Found STI sibling: Event (collection: events) from manifestCache:@happyvertical/smrt-events
```

This information could be used to merge subclass fields into the migration schema.

willgriffin:
## Summary

Add `smrtDependencies` discovery to the OXC scanner path, matching the TypeScript scanner path behavior from PR #707.

## Problem

The vite plugin has two scanner paths:
1. **TypeScript scanner** - Has `discoverExternalPackages: true` (PR #707)
2. **OXC scanner** (experimental) - Was **missing** smrtDependencies discovery

Packages using the OXC scanner (like caelus) had `smrtDependencies: null` in their manifests, breaking cross-package STI detection.

**Example:** WeatherForecast (caelus) extends Event (smrt-events). Event has `tableStrategy: 'sti'`, but without smrtDependencies, the runtime couldn't trace this relationship and generated a separate `weather_forecasts` table instead of merging into `events`.

## Changes

Added `discoverSmrtPackages()` call to `scanWithOxc()` function, populating `smrtDependencies` in the manifest.

## Verification

Tested with caelus rebuild - manifest now shows:
```json
{
  "smrtDependencies": [
    "@happyvertical/smrt-agents",
    "@happyvertical/smrt-core",
    "@happyvertical/smrt-events",
    "@happyvertical/smrt-places",
    "@happyvertical/smrt-profiles"
  ]
}
```

All 1173 tests pass.

Fixes #690

### DIFF
--- a/packages/core/src/vite-plugin/index.ts
+++ b/packages/core/src/vite-plugin/index.ts
@@ -534,6 +534,20 @@ export function smrtPlugin(options: SmrtPluginOptions = {}): Plugin {
       // Add moduleType identifier
       newManifest.moduleType = 'smrt';
 
+      // Discover external SMRT packages for cross-package STI inheritance
+      // (Issue #690: Without this, STI child classes from external packages
+      // get separate tables instead of merging into parent)
+      const { discoverSmrtPackages } = await import(
+        '../manifest/discover-smrt-packages.js'
+      );
+      const smrtDependencies = discoverSmrtPackages();
+      if (smrtDependencies.length > 0) {
+        newManifest.smrtDependencies = smrtDependencies;
+        console.log(
+          `[smrt] Found ${smrtDependencies.length} SMRT dependencies: ${smrtDependencies.join(', ')}`,
+        );
+      }
+
       // Generate pre-computed schemas for each object (same as TypeScript scanner path)
       const { ManifestGenerator } = await import('../scanner/index.js');
       const manifestGen = new ManifestGenerator();


willgriffin:
## CI Failure Analysis

The test failure is **not caused by this PR's code changes**, but rather exposes a pre-existing issue:

### Root Cause

1. smrt-core's vite plugin generates `.smrt/manifest.json` during dev/test builds
2. This manifest includes test classes (219 objects), including a `Secret` class from `cli-module.test.ts`
3. When smrt-secrets builds with the OXC scanner (this PR's fix), it now correctly discovers `@happyvertical/smrt-core` as a dependency
4. The manifest-loader loads smrt-core's `.smrt/manifest.json` (priority over `dist/manifest.json`)
5. smrt-secrets' own `Secret` class causes a collision

### Fix Required

Created issue #711 to rename smrt-secrets' `Secret` class to avoid collision.

**Alternative fixes:**
- Rename smrt-core's test `Secret` class in `cli-module.test.ts`
- Modify manifest-loader to prefer `dist/manifest.json` for external packages

This PR should wait for #711 to be resolved.

PR #882, opened 2026-02-09
feat(schema): support local STI subclasses in consuming apps

### LINKED ISSUE #881: feat(schema): Support local STI subclasses in consuming apps
## Problem

When a consuming app defines a local STI subclass (e.g., `Publication extends Tenant`), the schema generator doesn't know about it because:

1. The child class lives in the app, not in an upstream smrt package
2. The app has no `.smrt/manifest.json` — it relies on upstream package manifests
3. `smrt db:setup` and `smrt db:migrate` only read upstream manifests
4. From those manifests, the base class has no STI children, so `_meta_type` is never added to the table

## Example

```typescript
// In consuming app (e.g., blindmanpress dashboard)
import { Tenant } from '@happyvertical/smrt-users';

@smrt()
export class Publication extends Tenant {
  static readonly _meta_type = 'Publication';
  timezone: string = 'America/Edmonton';
  editorialEmail: string = '';
}
```

After `smrt db:setup`, the `tenants` table has no `_meta_type` column because the upstream `smrt-users` manifest doesn't include `Publication` in the STI hierarchy.

## Expected Behavior

Consuming apps should be able to register local STI subclasses so that:

- `smrt db:setup` creates the `_meta_type` column on the base table
- `smrt db:migrate` adds the column if missing
- STI indexes (`_meta_type`, `slug_context_meta_type`) are created

## Possible Approaches

1. **App-level manifest generation** — `smrt scan` runs on the app's local models and merges with upstream manifests
2. **Config-based registration** — `smrt.config.js` declares local STI extensions:
   ```js
   export default {
     sti: {
       'Tenant': ['./src/lib/models/Publication.ts']
     }
   }
   ```
3. **Runtime registration** — App registers STI children before schema operations:
   ```typescript
   ObjectRegistry.registerSTIChild('Tenant', Publication);
   ```

## Context

- The `tenants` table from `smrt-users` is a natural STI base — downstream apps commonly need typed tenant subtypes (Publication, Organization, etc.)
- All other STI hierarchies work because both base and children are in the same package manifest (Content → Article, Event → Meeting, etc.)
- This is blocking the blindmanpress dashboard seed script, which creates Publications as STI children of Tenant

willgriffin:
## Summary

- Add `generateAppManifest()` function that uses ManifestBuilder to scan app source files and generate `.smrt/manifest.json`
- Modify `autoDiscoverAndLoad()` to auto-scan when package manifests exist but no project manifest is found
- Add `smrt scan` CLI command for explicit app manifest generation
- Add tests for generateAppManifest and auto-scan behavior

## Problem

When a consuming app defines local STI subclasses (e.g. `Publication extends Tenant`), `smrt db:setup` and `smrt db:migrate` don't know about them because the app has no manifest. The vitest plugin already solves this for tests, but production CLI commands were missing this capability.

## Solution

Follow the same pattern as the vitest plugin: use ManifestBuilder + discoverBaseClasses to scan app source files and generate a project-level manifest. This happens automatically during `autoDiscoverAndLoad()` when no project manifest exists but package manifests are found.

## Test plan

- [x] `generateAppManifest()` returns null when no src/ directory exists
- [x] `generateAppManifest()` returns null when no SMRT objects found
- [x] `autoDiscoverAndLoad()` attempts auto-scan when package manifests exist but no project manifest
- [x] `autoDiscoverAndLoad()` skips auto-scan when project manifest already exists
- [x] `autoDiscoverAndLoad()` skips auto-scan when no package manifests exist
- [x] All 11 existing tests continue to pass

closes #881

### DIFF
--- a/packages/cli/src/__tests__/manifest-discovery.test.ts
+++ b/packages/cli/src/__tests__/manifest-discovery.test.ts
@@ -1,16 +1,16 @@
 /**
  * Tests for manifest discovery deduplication, version conflict detection,
- * and app-level manifest generation (issue #881)
+ * and missing manifest error (issue #881)
  */
 
-import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises';
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import {
   autoDiscoverAndLoad,
   discoverManifests,
-  generateAppManifest,
+  SmrtManifestNotBuiltError,
   SmrtVersionConflictError,
 } from '../discovery/manifest-discovery.js';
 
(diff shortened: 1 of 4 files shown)


willgriffin:
Closing — the root cause is in STI table resolution across package boundaries, not manifest discovery. The fix is adding `tableStrategy: 'sti'` to `Tenant` in smrt-users. See issue #881 comments for details.

PR #1070, opened 2026-03-23
smrt#1063: add owned asset joins for profiles, events, places, and products

willgriffin:
## Summary
- add dedicated owned-asset joins and helper APIs for profiles, events, places, and products
- export and document the new `*Asset` / `*AssetCollection` surfaces and extend the repo guidance for owned vs generic asset links
- add `ProductCollection` plus `smrt-products` library packaging hardening so the publish surface comes from `dist/lib`
- move the optional `users`/`tenancy` cross-package links to optional peer dependencies so Turbo accepts the workspace graph with the new profiles -> assets dependency

## Testing
- `pnpm -C packages/profiles exec vitest run src/__tests__/profile-assets.test.ts`
- `pnpm -C packages/events exec vitest run src/__tests__/event-assets.test.ts`
- `pnpm -C packages/places exec vitest run src/place-assets.test.ts`
- `pnpm -C packages/products exec vitest run src/product-assets.test.ts`
- `pnpm -C packages/content exec vitest run src/content-assets-migration.test.ts src/contents.spec.ts`
- `pnpm -C packages/images exec vitest run src/__tests__/deriver.test.ts`
- `pnpm -C packages/products run verify:pack`
- `pnpm run lint`
- `pnpm run typecheck`

### DIFF
--- a/packages/events/src/__tests__/event-assets.test.ts
+++ b/packages/events/src/__tests__/event-assets.test.ts
@@ -116,4 +116,23 @@ describe('Event owned assets', () => {
       [],
     );
   });
+
+  it('includes global assets when resolving tenant-owned event links', async () => {
+    const dbUrl = getTestDbUrl('event-assets-global');
+    const { event, assets } = await createEventFixture(dbUrl);
+
+    const globalAsset = await assets.create({
+      name: 'global-hero.jpg',
+      sourceUri: 'file:///tmp/global-hero.jpg',
+      mimeType: 'image/jpeg',
+      tenantId: null,
+    });
+    await globalAsset.save();
+
+    await event.addAsset(globalAsset, 'hero', 0);
+
+    expect((await event.getAssets('hero')).map((asset) => asset.id)).toEqual([
+      globalAsset.id,
+    ]);
+  });
 });
--- a/packages/events/src/models/Event.ts
+++ b/packages/events/src/models/Event.ts
@@ -6,7 +6,11 @@
 
 import type { Asset } from '@happyvertical/smrt-assets';
 import { SmrtObject, smrt } from '@happyvertical/smrt-core';
-import { TenantScoped, tenantId } from '@happyvertical/smrt-tenancy';
+import {
+  TenantScoped,
+  tenantId,
+  withSystemContext,
+} from '@happyvertical/smrt-tenancy';
 import type { EventOptions, EventStatus } from '../types';
 
 const ASSET_RELATIONSHIP_PATTERN = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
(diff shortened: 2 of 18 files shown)


willgriffin:
## PR Review: smrt#1070 — Owned Asset Joins

Overall this is a solid expansion of the owned-asset pattern to four more domain packages. Tests are thorough, docs are comprehensive, and the products packaging hardening is a welcome bonus. A few issues to flag:

---

### 🐛 Bug: `ProductAsset` missing `@TenantScoped` and `tenantId` field

`EventAsset`, `PlaceAsset`, and `ProfileAsset` all have:
```ts
@TenantScoped({ mode: 'optional' })
// ...
@tenantId({ nullable: true })
tenantId: string | null = null;
```
But `ProductAsset` has **neither** the decorator nor the field. This also means:
- `ProductAssetCollection.attach()` doesn't accept a `tenantId` parameter (compare the other three which all accept `tenantId: string | null = null`).
- `Product.addAsset()` doesn't forward `this.tenantId` to `attach()` (compare `Event.addAsset()`, `Place.addAsset()`, `Profile.addAsset()` which all pass `this.tenantId`).

If this is intentional because products aren't tenant-scoped, the inconsistency should be called out in CLAUDE.md. If it's an oversight, the fix is straightforward — add the decorator and plumb the `tenantId` through.

---

### 📐 Architecture: Significant code duplication across four packages

The following blocks are copy-pasted verbatim (modulo noun substitution) across events, places, profiles, and products:

1. **`assertValidAssetRelationship()` + `assertValidAssetSortOrder()`** — identical in 4 model files
2. **`resolveAssets()` / `getAssets()` / `addAsset()` / `removeAsset()`** on each model — ~70 identical lines per model
3. **`XAssetCollection`** — 93 near-identical lines per collection
4. **`XCollection` asset wrappers** — 34 identical lines per collection

This is ~800 lines of near-identical code. Consider extracting a shared mixin or base class in `@happyvertical/smrt-assets` (e.g., `AssetOwnerMixin` or an `OwnedAssetCollection` base) to eliminate the duplication. Even a shared util file for the validation helpers would reduce the surface.

Not a blocker for merge, but worth tracking as debt.

---

### ⚠️ Concern: `XAssetCollection` delegate methods create circular re-instantiation

Each `EventAssetCollection.getAssets()` / `addAsset()` / `removeAsset()` does:
```ts
const { EventCollection } = await import('./EventCollection');
const events = await EventCollection.create({ db: this.db });
return events.getAssets(eventId, relationship);
```
This creates a **new** `EventCollection` instance per call, which in turn creates a **new** `EventAssetCollection` inside the model's `getEventAssetCollection()`. The delegation creates unnecessary collection instantiation overhead. Consider either:
- Having `XAssetCollection` implement the logic directly (inline the model lookup + `resolveAssets`)
- Or caching the delegate collection instance

Same pattern in all four `*AssetCollection` classes.

---

### ⚠️ Products `package.json` exports collapse all sub-paths to the root barrel

All of `./models`, `./components`, `./stores`, `./generated`, `./utils`, and the new `./collections` now point to `./dist/lib/index.js`:
```json
"./models": { "default": "./dist/lib/index.js" },
"./components": { "default": "./dist/lib/index.js" },
```
This means `import { Category } from '@happyvertical/smrt-products/models'` works but gives consumers the entire package tree, defeating the purpose of sub-path exports. Existing consumers that relied on targeted imports (e.g. `./components` for just Svelte components) will now pull in all models, collections, and server code.

If the intent is to simplify the export surface temporarily, a code comment or CLAUDE.md note would help future maintainers understand why.

---

### 🔧 Nit: `vite.config.ts` indentation drift in products

The `return` statement got an extra tab of indentation:
```diff
-	return {
+		return {
```
This shifts the entire block one level deeper without a structural reason.

---

### ✅ Things done well
- Tests cover model API, conflict/upsert dedup, and collection-level wrappers for all four packages
- Consistent use of `conflictColumns` for idempotent upserts
- Clean documentation updates across AGENTS.md, CLAUDE.md, and README.md files
- Smart move of `users`/`profiles` cross-deps to optional peer dependencies to unblock the workspace DAG
- Products build script actually building now instead of skip-echoing

PR #1129, opened 2026-04-18
feat(assets): public asset runtime + generic serving contract (#1128)

### LINKED ISSUE #1128: smrt-assets: expose a shared agent asset runtime and generic serving contract
## Summary

`@happyvertical/smrt-assets` already has the right data model for multi-tenant assets, but the runtime/serving contract is still too app-specific.

In `anytown.ai` we now have two real agent use cases that both need first-class asset access:

- `Praeco` source documents (agendas/minutes/highlights) and upcoming extracted document images
- `Histrio` generated media assets

Today the asset *records* are shared, but the asset *runtime* is not:

- Histrio has a custom `HistrioDb` wrapper that owns `AssetCollection` + `AssetStore`
- Dashboard asset-serving routes are Histrio-shaped
- Praeco can create/link `Asset` records, but does not have the same explicit asset-store/runtime surface for derived assets

That means `smrt-assets` works as a model package, but not yet as a first-tier base package for real apps with system assets, network-tenant assets, site-tenant assets, and object-owned derived assets.

## Problem

We need a standard upstream contract for agents/apps to:

1. get an `AssetStore` / asset runtime without inventing app-specific wrappers
2. create source assets and derived assets consistently
3. serve/read assets through a generic app-level path that is not tied to one agent
4. support tenant-scoped and global assets uniformly

Without that, every app ends up re-inventing the same glue, or worse, reaching into another agent's private plumbing.

### Runtime

A public, documented way for an app or agent to get a shared asset runtime, for example:

- `AssetCollection`
- `AssetStore`
- storage provider config
- optional derived-asset helpers

### Agent integration

A clean pattern for agent options / constructors to accept the shared asset runtime, similar to how agents already accept shared DB options.

### Serving

A generic serving/read contract for `smrt-assets`-backed assets that apps can wire once, instead of Histrio-specific routes becoming the de facto standard.

### Relationships / conventions

Document shared conventions for:

- source asset vs derived asset
- parent/child derivatives
- relationship labels like `source_document`, `document_image`, `thumbnail`, `proof`
- metadata for extraction status, source URL, hashes, page numbers, etc.

## Why this matters now

Anytown is about to standardize on `smrt-assets` as a first-tier base package for:

- system assets
- network-tenant assets
- site-tenant assets
- object-owned assets across multiple agents

The model layer is already close. The missing piece is the runtime/serving contract.

## Notes

This is not a request to make `smrt-assets` a dependency of `smrt-core`. It should stay separate. The ask is to make it a stronger first-tier package for real app stacks by giving it the missing public runtime surface.

willgriffin:
## Summary

Closes #1128. Gives `@happyvertical/smrt-assets` a first-tier public runtime / serving contract so apps and agents (Praeco, Histrio, etc.) can stop inventing bespoke wrappers around `AssetCollection` + `AssetStore`.

- `AssetRuntime` bundles `AssetCollection`, `AssetAssociationCollection`, and `AssetStore` behind one handle. Factory: `createAssetRuntime({ db, storage })`.
- `storeSourceAsset` / `storeDerivedAsset` / `linkDerivation` / `setExtractionStatus` cover the common source-document → derivative flows, including provenance `AssetAssociation`s.
- `serveAsset()` + `resolveAssetForServing()` give apps a framework-agnostic Web `Response` (or bytes + headers) with built-in tenant and `canAccess` checks. Works in SvelteKit `+server.ts`, Hono, plain Node 18+, or custom runtimes via `responseCtor`.
- `ASSET_ROLES` / `ASSET_METADATA_KEYS` / `ASSET_EXTRACTION_STATUS` export the shared vocabulary (`source_document`, `document_image`, `thumbnail`, `proof`, `derivation_source`, ...) and extraction-state keys described in the issue.
- `packages/assets/CLAUDE.md` documents the runtime surface, the serving contract, and the shared role/metadata conventions.

## Test plan

- [x] `pnpm --filter @happyvertical/smrt-assets build` — clean build with manifest generation
- [x] `pnpm --filter @happyvertical/smrt-assets test` — 20/20 tests pass (including 18 new ones across `asset-runtime.test.ts` and `asset-serving.test.ts`)
- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `turbo build --filter=@happyvertical/smrt-images --filter=@happyvertical/smrt-content` — downstream packages still build
- [ ] Reviewer sanity-check that the conventions match the anytown.ai Praeco/Histrio use cases

### DIFF
--- a/packages/assets/src/asset-runtime.ts
+++ b/packages/assets/src/asset-runtime.ts
@@ -97,20 +97,25 @@ export interface StoreDerivedAssetOptions
 
   /**
    * If true (default), also create an `AssetAssociation` record with
-   * `assetId=source.id`, `metaType=<sourceMetaType>`, `metaId=<newId>`,
-   * `role=<role>` so the provenance link is queryable independent of
-   * `parentId`.
+   * `assetId=source.id`, `metaType=<derivativeMetaType>`,
+   * `metaId=<newAssetId>`, `role=<role>` so the provenance link is
+   * queryable independent of `parentId`.
    *
    * Set to `false` if you only want the hierarchical `parentId` link.
    */
   linkAssociation?: boolean;
 
   /**
-   * `metaType` string to use when creating the association. Defaults to
-   * `'Asset'`. Callers with STI subclasses (e.g. `Image`) should pass
-   * the subclass name here.
+   * `metaType` string stored on the `AssetAssociation`. This describes
+   * the *derivative* object (the target of `metaId`), not the source.
+   *
+   * Defaults to `'Asset'`. Callers whose derivatives are STI subclasses
+   * — e.g. an `Image` produced from a PDF — should pass the subclass
+   * name here so consumers can distinguish subtype derivatives from
+   * plain assets. This matches the convention used by
+   * `smrt-images`' `ImageDeriver.deriveWithAssociations()`.
    */
-  sourceMetaType?: string;
+  derivativeMetaType?: string;
 }
 
 /**
@@ -118,7 +123,11 @@ export interface StoreDerivedAssetOptions
  */
 export interface LinkDerivationOptions {
   role?: AssetRole | string;
-  sourceMetaType?: string;
+  /**
+   * `metaType` describing the derivative object (target of `metaId`).
+   * See `StoreDerivedAssetOptions.derivativeMetaType`.
+   */
+  derivativeMetaType?: string;
 }
 
 /**
@@ -171,7 +180,7 @@ export class AssetRuntime implements AssetRuntimeLike {
     const {
       role: rawRole,
       linkAssociation = true,
-      sourceMetaType = 'Asset',
+      derivativeMetaType = 'Asset',
       ...storeOpts
     } = opts;
     const role = rawRole ?? ASSET_ROLES.DERIVATION_SOURCE;
@@ -184,7 +193,7 @@ export class AssetRuntime implements AssetRuntimeLike {
     if (linkAssociation && derived.id) {
       await this.associations.associate(
         source.id,
-        sourceMetaType,
+        derivativeMetaType,
         derived.id,
         role,
       );
@@ -208,10 +217,10 @@ export class AssetRuntime implements AssetRuntimeLike {
       );
     }
     const role = opts.role ?? ASSET_ROLES.DERIVATION_SOURCE;
-    const sourceMetaType = opts.sourceMetaType ?? 'Asset';
+    const derivativeMetaType = opts.derivativeMetaType ?? 'Asset';
     return this.associations.associate(
       source.id,
-      sourceMetaType,
+      derivativeMetaType,
       derivative.id,
       role,
     );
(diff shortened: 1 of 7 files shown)


packages/assets/src/asset-runtime.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    const derived = await this.store.store(name, data, {
+      ...storeOpts,
+      parentId: source.id,
+    });
+
+    if (linkAssociation && derived.id) {
+      await this.associations.associate(
+        source.id,
+        sourceMetaType,
```
willgriffin: Unless I'm missing a different `AssetAssociation` convention here, this is recording the wrong `metaType`. `metaId` is the derivative asset id, so `metaType` needs to describe the derivative object (`Asset`, `Image`, etc.), not the source. In Anytown we need these derivation links to distinguish plain document assets from image/page derivatives, and this currently records subtype derivatives as the source type. The existing `smrt-images` path uses `associate(source.id!, 'Image', derived.id!, 'derivation-source')`, which matches the current collection semantics.

packages/assets/src/asset-serving.ts
```diff
@@ -0,0 +1,240 @@
...
+  if (canAccess) {
+    const allowed = await canAccess(asset);
+    if (!allowed) {
+      throw new AssetServeError('Asset access denied', 403);
+    }
+  }
+
+  let data: Buffer;
+  try {
+    data = await runtime.store.read(asset);
```
willgriffin: This currently only works for assets whose bytes live behind `AssetStore.read()`. In Anytown we already have a real mixed fleet: Histrio serves locally stored assets, but Praeco source-document assets are often `http(s)` `sourceUri`s that we intentionally proxy directly from the origin document URL. Our shared dashboard asset runtime has an explicit remote-asset branch for that case. With the helper as written, swapping to `serveAsset()` would turn those document assets into 500s even though they are valid assets today. I think `resolveAssetForServing()` needs the same remote-URI handling before this is a true generic serving contract for us.

willgriffin (commented):
I pulled this against Anytown's current Praeco/Histrio asset usage. The public runtime direction is the right one, but I found two integration blockers for us before we can replace our downstream wrappers cleanly.

willgriffin:
Re-reviewed against Anytown after the update. The two original blockers for us are fixed now:

- derivation associations now record the derivative `metaType`
- `serveAsset()` / `resolveAssetForServing()` now handle remote `http(s)` `sourceUri`s, which is the shape Praeco uses for source-document assets today

Focused package tests also pass locally on the updated head:
- `pnpm --dir packages/assets exec vitest run src/__tests__/asset-runtime.test.ts src/__tests__/asset-serving.test.ts`

The one remaining adoption caveat on our side is `setExtractionStatus()`: it intentionally treats `asset.description` as a JSON sidecar, which would overwrite the human-readable source-document descriptions Praeco already stores unless we avoid that helper or move that prose into a reserved JSON key first. I don’t see that as a blocker for merging this PR, but it is something we’d need to account for before wiring extraction-state writes in Anytown.

willgriffin:
Thanks for the re-review! Pushed [e680908b](https://github.com/happyvertical/smrt/pull/1129/commits/e680908b) to address the adoption caveat too:

`setExtractionStatus()` is now non-destructive when `asset.description` is free-form prose — the original value is preserved under a reserved `text` key of the sidecar instead of being overwritten. Existing JSON-object descriptions are merged into as before. So Praeco can wire extraction-state writes against its current source-document descriptions without a migration step.

Two new tests cover the prose-preservation + JSON-object-merge paths; 32/32 green locally. CLAUDE.md updated too.

Issue #1134, opened 2026-04-21
refactor(core): decision-table register() + strict diagnostics default + retire rehydration flag


willgriffin:
## Summary

Third and final release of the ObjectRegistry / SmrtObject review (parent thread: #1132). Two user-visible behavior changes + one internal cleanup:

1. **Decision-table rewrite of `register()`** in `packages/core/src/registry/class-registration.ts` — collapse the 11-branch exact-match + case-insensitive collision policy, plus `registerFromManifest`'s parallel 3-branch `resolveManifestCollision`, into a single `decideCollisionPolicy()` function with an auditable decision table.
2. **Flip `SMRT_STRICT_REGISTRY` to default-on** — the opt-in strict-mode error path added in Release A becomes the default. Opt-out via `SMRT_STRICT_REGISTRY=false`.
3. **Retire `SMRT_SKIP_STI_REHYDRATE`** — delete the env-flag opt-out added in Release A. The unconditional rehydration behavior (current default since #1131) becomes the only behavior; the flag disappears from the public surface. Further optimizing away the rehydration loop itself is deferred to a follow-up with its own regression coverage.

Both Release A ([#1135](https://github.com/happyvertical/smrt/pull/1135)) and Release B ([#1138](https://github.com/happyvertical/smrt/pull/1138)) are merged.

> **Plan revisions from Codex review** (see [comment](https://github.com/happyvertical/smrt/issues/1134#issuecomment-4290823035)):
> 1. Scope of §4 narrowed — flag deleted, rehydration loop stays; the `external-runtime-hydration.test.ts:1071-1202` stale-cache repair contract is preserved rather than regressed
> 2. `merge-manifest` policy's alias-promotion step documented as caller responsibility (stays pure in the decision function)
> 3. Decision table's invocation contract made explicit: only called when the caller has identified exactly one colliding candidate. Ambiguous simple-name matches (issue #951 coexistence) fall through without consulting the table
> 4. Diagnostic call-site count corrected from seven to six; test audit enumerated for the strict-mode flip

---

### 1.1 Exact-match path ([class-registration.ts:161-278](packages/core/src/registry/class-registration.ts:161))

| # | Condition | Action | Lines |
|---|---|---|---|
| 1 | Same constructor, re-registration | accept (`upsertExistingEntry`) | [170-173](packages/core/src/registry/class-registration.ts:170) |
[...]

Issue #1133, opened 2026-04-21
refactor(core): consolidate ObjectRegistry manifest discovery + drop redundant index


willgriffin:
## Summary

Follow-up to the #1132 review. Collapse ObjectRegistry's four overlapping manifest-discovery paths into a single `ManifestSource` interface, and drop the redundant `classNameMap` index so there's only one canonical way to reach a registered class. Internal refactor only — the public `ObjectRegistry.*` method surface stays stable.

Release A ([#1135](https://github.com/happyvertical/smrt/pull/1135)) is now merged: every `@happyvertical/smrt-*` package self-registers its manifest at import time via a `__smrt-register__.ts` shim that calls `ObjectRegistry.registerPackageManifest()`. That's the precondition Release B builds on — once packages reliably seed `__smrtManifestCache` before their `@smrt()` decorators fire, the on-demand fs-walking paths become a fallback rather than the primary discovery strategy.

> **Plan revisions from Codex review** (see [comment](https://github.com/happyvertical/smrt/issues/1133#issuecomment-4289887745)): preserved `loadAllManifests({ manifestPaths })` as a first-class explicit-paths source, made the lookup API qualified-name/package-aware, kept the current composite priority order (Static before Embedded), added a preliminary commit that extracts a leaf `manifest/store.ts` module to break the existing `registry ↔ manifest-loader` cycle before moving helpers, and resequenced so `classNameMap` removal lands **last** after discovery parity is proven.

---

## 1. Current discovery surface (what Release B is collapsing)

Current state: **four overlapping discovery paths** distributed across two files, reading/writing **nine globalThis caches**, with **six read sites** for the `classNameMap` index.

### 1.1 Discovery entry points

| Path | File | Line | Purpose | Sync/Async |
|---|---|---|---|---|
| `registerPackageManifest(url)` | registry.ts | [1522](packages/core/src/registry.ts:1522) | Seeds `__smrtManifestCache` at package import (Release A) | sync |
| `tryLoadFromExternalPackage(className)` | registry.ts | [1229](packages/core/src/registry.ts:1229) | Walks `node_modules` to find a class's manifest; calls `registerFromManifest` | async |
[...]

PR #1135, opened 2026-04-21
Release A: self-registering manifests (closes #1132) + diagnostics + STI rehydration flag

### LINKED ISSUE #1132: Consumer-runtime ObjectRegistry: declared fields dropped when smrt packages are consumed from node_modules (non-vitest runtimes)
## Summary

When a smrt package (e.g. `@happyvertical/smrt-places`) is consumed from an application's `node_modules` at runtime, its models' **declared fields aren't registered in `ObjectRegistry`**, so `.toJSON()` serializes only the base `SmrtObject` columns. Everything else (`name`, `latitude`, `externalId`, …) is silently dropped on `save()`, and `list({ where: { … } })` validation rejects those same fields as unknown.

This is the runtime analogue of #583 (which was fixed for test environments by `smrtVitestPlugin` / `ManifestManager`). The test-side fix works — `smrt-places`' own `place-assets.test.ts` round-trips fields correctly. But there's no equivalent registration step for non-vitest consumer runtimes (Node/tsx scripts, SvelteKit SSR, plain `vite dev`, etc.).

## Repro

Running the recently-merged `PlaceCollection.discoverNearby` / `resolveTrackPlaces` from the `feat/places-poi-discovery` branch against real Overpass POIs, through a consumer app that imports `@happyvertical/smrt-places` via pnpm-linked sibling `node_modules`:

1. Consumer calls `await places.create({ name: 'Guru Kitchen + Bar', latitude: 53.546, longitude: -113.494, externalId: 'osm-node-10958146781', source: 'openstreetmap', … })` → `await place.save()`.
2. Postgres row inserted with `id`, `slug`, `context`, `_meta_type`, timestamps — **everything else empty**:
   ```
   SELECT id, name, latitude, longitude, external_id, _meta_data FROM places;
      id                                  | name | latitude | longitude | external_id | _meta_data
      87078ad4-ff79-48a8-945e-1d777ee7dfeb |      |          |           |             | {}
      52a80327-05e8-4d91-9cc9-d1c42f51fbe9 |      |          |           |             | {}
   ```
   (Even `_meta_data` is `{}` — the STI meta path isn't catching them either.)
3. Consumer calls `await places.list({ where: { externalId: 'osm-node-10958146781' } })` →
   ```
   Invalid WHERE clause field: 'externalId'. Field does not exist on Place.
     Valid fields: _meta_data, _meta_type, context, created_at, id,
                   meta_data, meta_type, slug, tenant_id, updated_at
       at PlaceCollection.convertWhereKeys (…/core/dist/collection.js:128:15)
   ```
   The \"valid fields\" list is exactly the base `SmrtObject` field set — Place's 22 declared fields aren't in `ObjectRegistry`.

Same behaviour reproduces for `Asset` (`size_bytes` / `duration_ms`) when consumers touch fields added via raw migration outside the decorator's view.

## Why it matters

`PlaceCollection.lookupOrCreate`, `findByCoordinates`, `findByQuery`, and the new `discoverNearby` / `resolveTrackPlaces` all rely on `list({ where: { latitude: … } })` or `list({ where: { externalId: … } })` and on `create(...) → save()` round-tripping declared fields. Every one of them silently no-ops (or throws \"Invalid WHERE clause field\") from a consumer runtime that hasn't manually loaded the manifests.

In the ergot.io reference project this blocks the full end-to-end demonstration of the POI discovery pipeline that just landed across `@happyvertical/geo`'s `findPoisNear` PR and this package's `discoverNearby` / `resolveTrackPlaces` PR. The data-plane side (provider calls, bucketing, throttling, association inserts) works; the Place rows it creates round-trip empty, so the UI shows \"Unnamed place\" and no map coordinates.

## Diagnosis

Three confirming facts:

- `@happyvertical/smrt-places/dist/manifest.json` contains all 22 Place fields with correct types. Codegen is fine.
- `place-assets.test.ts` passes because `smrtVitestPlugin` invokes `ManifestManager` to load manifests from peer deps before tests run (`packages/vitest/src/index.ts:5-8`).
- Consumers that don't use that plugin have no equivalent step. `ObjectRegistry.getFields('Place')` returns only the base class fields at runtime.

So the root cause is **there is no canonical runtime step to load a consumed package's manifest into `ObjectRegistry`** outside of vitest.

## Expected behaviour

When an app does `import { PlaceCollection } from '@happyvertical/smrt-places'`, Place's declared fields should end up in `ObjectRegistry` before the first `save()` / `list({ where })` so that:

- `.toJSON()` writes all declared fields.
- Where-clause validation accepts them as valid column names.

Ideally this happens automatically (module-side effect on import) without the consumer needing to wire a plugin.

## Possible approaches

1. **Import-time self-registration**: have the smrt package's entry point (`dist/index.js`) call something like `ObjectRegistry.registerFromManifest(manifestJson)` on module load, using its own bundled `manifest.json`. Most portable — works in every runtime with no plugin.
2. **Consumer runtime helper**: ship a `smrtConsumer` (or extend the existing Vite `smrt-core/consumer-plugin`) that walks `node_modules/@happyvertical/*/dist/manifest.json` the way `smrtVitestPlugin` does and preloads them into `ObjectRegistry`. Explicit, but needs to be wired by every app.
3. **Decorator-time registration for packaged consumers**: make the `@smrt` decorator self-register at class-definition time by reading package-bundled metadata emitted at build. Removes the manifest-file dependency entirely but likely requires codegen changes.

My lean is **(1)** — the smallest per-package change, no consumer opt-in, survives Node/tsx/vite/SvelteKit alike. Happy to help if it's useful.

## Scope

Affects every smrt-\\* package with declared fields consumed from node_modules in non-vitest runtimes. We hit it with `smrt-places` and have a side-channel workaround for `smrt-assets` extras (`size_bytes`, `duration_ms`) that I'd love to retire once this lands.

## Cross-refs

- Related prior fix: #583 (vitest-side manifest loading).
- Downstream consumer: `willgriffin/ergot.io` feat/poi-discovery — the POI + track resolution UI lights up the moment Place's fields round-trip.

willgriffin:
## Summary

Release A of the ObjectRegistry / SmrtObject review. Closes #1132 (consumer-runtime field drop) by making every `@happyvertical/smrt-*` domain package self-register its build-time manifest at import time, before any `@smrt()` decorator fires.

Previously only the vitest plugin populated the global manifest cache; tsx scripts, SvelteKit SSR, plain `vite dev`, and any other non-vitest runtime silently registered classes with zero declared fields — `place.save()` wrote only base SmrtObject columns and `list({ where: { externalId } })` rejected the field as unknown.

## What changed

**`@happyvertical/smrt-core`**
- `ObjectRegistry.registerPackageManifest(url)` — reads a manifest JSON URL, seeds `__smrtManifestCache` under the package key, and merges manifest fields into any class a decorator already registered. Silent no-op if the URL can't be read (dev/test paths hit other sources).
- `ObjectRegistry.getDiagnostics() / clearDiagnostics() / flushDiagnostics()` — opt-in collector for the 3 silent-failure sites in `registerPackageManifest` / `resolveManifestExportPathSync` / `resolveManifestExportPathSync` (`MANIFEST_EXPORT_INVALID`, `MANIFEST_EXPORT_NOT_JSON`, `MANIFEST_EXPORT_NOT_FOUND`, `PACKAGE_MANIFEST_READ_FAILED`, `PACKAGE_MANIFEST_INVALID_SHAPE`). Passive by default; `SMRT_STRICT_REGISTRY=true` makes severity-`error` diagnostics throw at record time. Release C (#1134) flips strict on by default.
- `SMRT_SKIP_STI_REHYDRATE=true` — opts out of #1131's unconditional STI descendant re-hydration on every save. With self-registration in place those caches aren't stale to begin with, but #1131's behavior stays default for safety. Flag removed in Release C once rollout is proven.

**27 domain packages** (ads, affiliates, agents, analytics, assets, chat, commerce, content, events, facts, features, images, jobs, ledgers, messages, places, products, profiles, projects, properties, secrets, sites, social, tags, tenancy, users, video, voice)
- Each gains a `src/__smrt-register__.ts` one-liner that calls `ObjectRegistry.registerPackageManifest(new URL('./manifest.json', import.meta.url))`.
- Each `src/index.ts` imports that file as its first statement so the side effect runs before any class module evaluates.

## Breaking changes

None. All existing tests pass. Public API is strictly additive.

## Follow-ups

- [#1133](https://github.com/happyvertical/smrt/issues/1133) — Release B: collapse the 4 overlapping manifest discovery paths into a single `ManifestSource` interface; drop redundant `classNameMap` index. Internal refactor, no public-API change.
- [#1134](https://github.com/happyvertical/smrt/issues/1134) — Release C: decision-table rewrite of `register()`, flip `SMRT_STRICT_REGISTRY` on by default, retire `SMRT_SKIP_STI_REHYDRATE` flag (delete the unconditional rehydration code).

## Verification

- New unit tests: `packages/core/src/__tests__/register-package-manifest.test.ts` (6 tests — hydration before/after decorator, silent no-ops for missing/malformed manifests, string-path and URL input).
- `npx turbo build` — 41/41 packages build cleanly. Verified that `ObjectRegistry.registerPackageManifest(...)` is inlined at the top of each built `dist/index.js` ahead of all class decorator calls.
- `npx turbo test` — 80/80 test tasks successful.

## Test plan

- [ ] Verify #1132 reproducer: `pnpm add @happyvertical/smrt-places` into a fresh tsx script, create + save + list a Place, confirm all declared fields round-trip.
- [ ] Verify SvelteKit SSR runtime: import from `@happyvertical/smrt-places` in a `+page.server.ts`, confirm fields populate.
- [ ] Smoke-test `ObjectRegistry.getDiagnostics()` returns an empty array on a healthy startup.
- [ ] Verify `SMRT_STRICT_REGISTRY=true` throws when a malformed manifest is fed to `registerPackageManifest`.
- [ ] Confirm `SMRT_SKIP_STI_REHYDRATE=true` reverts to the pre-#1131 conditional hydration path (existing STI tests still pass under the flag).


packages/jobs/src/index.ts
```diff
@@ -32,6 +32,11 @@
  * ```
  */
 
+// Self-register this package's manifest before any @smrt() decorator fires
+// downstream. Must come first so the side effect runs ahead of the class
+// module loads below. See __smrt-register__.ts for issue #1132 context.
+import './__smrt-register__';
```
willgriffin: [P1] This only protects the package when consumers import the root entry. We still publish model-bearing subpaths such as `@happyvertical/smrt-jobs/runner`, `@happyvertical/smrt-products/models` / `collections`, and `@happyvertical/smrt-secrets/models` / `service`, and those entrypoints never execute `__smrt-register__`. A consumer that imports one of those supported subpaths can still hit the zero-field registration bug this PR is meant to close.

packages/core/src/registry.ts
```diff
@@ -1461,6 +1486,150 @@ export class ObjectRegistry {
...
+    if (!builtins) {
+      return { loaded: false, objectsRegistered: 0 };
+    }
+
+    let manifest: SmartObjectManifest | null = null;
+    try {
+      const urlString =
+        manifestUrl instanceof URL ? manifestUrl.href : manifestUrl;
+      const filePath = urlString.startsWith('file:')
+        ? builtins.path.normalize(new URL(urlString).pathname)
```
willgriffin: [P2] `URL.pathname` keeps encoded characters like `%20` and also yields `/C:/...`-style paths on Windows. That means `existsSync(filePath)` can fail for a valid `file:` URL and we silently return without recording any diagnostic. This should use `fileURLToPath()` and emit a not-found diagnostic before returning.

willgriffin:
Thanks for the reviews — all four substantive comments addressed in bd3331bf:

**P1 — Windows/URL decoding (chatgpt-codex, Copilot, willgriffin)**
`registerPackageManifest` now uses `fileURLToPath()` for `file:` URLs instead of `new URL(...).pathname`, fixing both Windows drive-letter paths (`/C:/foo`) and percent-encoded segments (spaces, unicode). Added a `PACKAGE_MANIFEST_NOT_FOUND` diagnostic so a path that fails `existsSync` is no longer a silent no-op — it shows up in `ObjectRegistry.getDiagnostics()`. Regression test in `register-package-manifest.test.ts:159` exercises a manifest path under `dir with spaces/`.

**P1 — Subpath entries bypass registration (willgriffin)**
Rolled the `__smrt-register__.js` import into every subpath entry source across the 28 packages — 52 files in total, including the ones called out (`jobs/runner`, `products/{models,collections,...}`, `secrets/{models,service}`) plus all the UI/svelte/playground/vite/server/adapters/testing/utils entries. The `registerPackageManifest` call is idempotent (second hit is a cache lookup), so the cost is ~0 if a consumer also imports the main entry.

**.js specifier consistency (Copilot)**
Every shim import is now `import './__smrt-register__.js';` to match the rest of the source.

**Unused `mkdirSync` (Copilot)**
Addressed — though the test now uses `mkdirSync` in the new "decodes percent-encoded file URLs" regression test, so the import remains.

Full test suite still green: 80/80 turbo test tasks pass, 41/41 packages build, 7/7 unit tests on `registerPackageManifest` pass.

willgriffin:
CI failure in d190d42e was the overly-broad subpath rollout from the previous review fix: the shim was added to every subpath entry including UI-only / node-only ones (`ui.ts`, `svelte/index.ts`, `playground.ts`, `vite-plugin.ts`, `server/index.ts`, `adapters/`, `testing.ts`, `utils.ts`, `sveltekit.ts`, `mock-smrt-client.ts`, `components/`, `stores/`).

`smrt-playground-host`'s Vite build aliases every `@happyvertical/smrt-*` to `src/` for the browser bundle. The shim import chains from those subpaths transitively pulled smrt-core's manifest generator into the client bundle, which dynamically loads the scanner via `tsx` — Rollup then tried to bundle `tsx` and failed on `fsevents`.

Narrowed the rollout to just the subpaths that actually re-export `@smrt()`-decorated classes and match the review's concrete call-outs:

- `jobs/runner`
- `products/{models, collections, generated}`
- `secrets/{models, services/SecretService}`

UI / Svelte / playground / vite / server / adapters / testing / utils entries either don't instantiate `@smrt()` classes at import time or are node-only (never ship to the browser), so they don't need the shim.

Local verification: `smrt-playground-host#build` now passes; `npx turbo build` 41/41 green, `npx turbo test` 80/80 green. Watching CI.

Issue #1139, opened 2026-04-21
perf(core): eager inheritedFields invalidation to retire STI save-time rehydration loop


willgriffin:
## Summary

Follow-up to [#1134](https://github.com/happyvertical/smrt/issues/1134) (Release C). Remove the save-time STI rehydration loop in `packages/core/src/object.ts` and replace it with eager `inheritedFields` invalidation on re-registration.

This was considered and explicitly **deferred out of Release C** because the correctness contract is non-trivial — see [Codex review](https://github.com/happyvertical/smrt/issues/1134#issuecomment-4290823035) for the analysis. Release C retires just the `SMRT_SKIP_STI_REHYDRATE` env flag; the rehydration loop itself stays as the single behavior.

## Problem

`object.ts:1134-1172` calls `ObjectRegistry.getAllFields()` on **every** STI descendant (including the saving class itself) on **every** `save()` call. In a hierarchy with N descendants, that's N async calls per save. For a table like `smrt-content` / `smrt-video` with tens of STI members, this is the dominant cost of `save()` for small rows.

## Why it's there

PR #1131 added the loop to fix [#1132](https://github.com/happyvertical/smrt/issues/1132) (stale `inheritedFields` caches in consumer runtimes where Release A's self-registration hadn't yet landed). Release A fixed the root cause; Release B's manifest-source abstraction made descendant registrations reliable; Release C (#1134) retires the opt-out flag. But the loop itself still runs unconditionally because the [`external-runtime-hydration.test.ts:1071-1202`](packages/core/src/__tests__/external-runtime-hydration.test.ts:1071) regression test pins down a "stale-but-present `inheritedFields` repair" contract that the conservative (pre-#1131) hydration path would regress.

## Proposed approach

Replace save-time repair with **eager invalidation on re-registration**. Every code path that can produce a stale-but-present `inheritedFields` state must invalidate cleanly:

1. **Audit the stale-producing paths** — walk every `classes.set(...)` and `existing.fields = ...` / `existing.inheritedFields = ...` site. Each must call `invalidateInheritanceEntries(existing)` after mutation.
2. **Prove correctness** — the synthetic stale state in `external-runtime-hydration.test.ts:1071` is seeded manually (`imageRegistered!.inheritedFields = new Map(imageRegistered?.fields)`). Identify whether any production code path can reach the same shape. If yes, invalidate there; if no, rewrite the test to exercise a realistic scenario.
[...]

Issue #1000, opened 2026-03-08
bug: db:migrate does not create advertising tables (customers, contracts, contract_line_items)


willgriffin:
## Description

`smrt db:migrate` does not create the `customers`, `contracts`, or `contract_line_items` tables that are referenced by the advertising module (`@happyvertical/suasor`) via raw SQL queries.

## Reproduction

1. Fresh database with `smrt db:migrate` applied
2. Dashboard overview calls `getDashboardStats()` which queries `SELECT COUNT(*) as count FROM customers WHERE status = 'active'`
3. Fails with: `relation "customers" does not exist, code=42P01, severity=ERROR`

## Expected Behavior

`smrt db:migrate` should create all tables required by registered SMRT packages, including tables used by `@happyvertical/suasor`'s advertising features (customers, contracts, contract_line_items).

## Affected Queries

The following raw SQL in `apps/dashboard/src/lib/server/advertising.ts` references these missing tables:

- `customers` — advertiser profiles, credit limits, payment terms, status
- `contracts` — advertiser contracts with references, amounts, terms
- `contract_line_items` — line items within contracts

## Environment

- smrt v0.20.47
- PostgreSQL 17 (Docker)
- anytown.ai dashboard app

## Workaround

None currently — the advertising dashboard sections fail silently (caught errors) but log `DATABASE_ERROR` on every overview page load.

Issue #951, opened 2026-02-20
Registry: use package-qualified names as primary keys


willgriffin:
## Summary

Follow-up to #950. The `ObjectRegistry` uses a flat `Map<string, string>` (`classNameMap`) keyed by lowercase simple name. This means only **one class per simple name** can exist in the registry — the second is either silently dropped (`registerFromManifest()`) or triggers a collision error (`register()`).

#950's Option B (inheritance-aware `registerFromManifest()`) handles the STI case, but doesn't solve the general problem: two genuinely unrelated packages defining a class with the same simple name. As the ecosystem grows (more vertical agents, more domain packages), simple-name collisions between unrelated classes will become increasingly common.

## Proposal

Make **package-qualified names** (e.g., `@happyvertical/smrt-events:Event`) the primary identity of registered classes. Simple names become **lookup aliases** that resolve when unambiguous.

### Core Data Structure Change

```typescript
// BEFORE: Map<string, string> — one entry per simple name
classNameMap: "event" → "@happyvertical/smrt-events:Event"

// AFTER: Map<string, string[]> — multiple entries per simple name
classNameMap: "event" → [
  "@happyvertical/smrt-events:Event",
  "@happyvertical/smrt-conferences:Event"
]
```

### Lookup Semantics

- **Qualified name lookup** (`findClass("@happyvertical/smrt-events:Event")`) — direct Map hit, always unambiguous
- **Simple name lookup** (`findClass("Event")`) — resolves the alias array:
  - Length 1 → return it (common case, zero friction)
  - Length > 1 → disambiguate by caller context (package name hint, tenant config) or surface an actionable error: *"'Event' is ambiguous — did you mean @happyvertical/smrt-events:Event or @happyvertical/smrt-conferences:Event?"*

### Affected Code Paths

These all need to be updated to work with qualified-name primary keys:

| Location | Current Behavior | Required Change |
|----------|-----------------|-----------------|
| `classNameMap` getter (`registry.ts:810`) | `Map<string, string>` | `Map<string, string[]>` |
| `getCanonicalClassName()` (`registry.ts:822`) | Returns single string | Returns array, caller disambiguates |
| `registerFromManifest()` (`registry.ts:1751`) | Silent early return on name match | Always register under qualified key; append to alias array |
| `register()` (`registry.ts:992`) | Collision detection via flat key | Register under qualified key; STI child-wins still applies |
[...]

PR #1140, opened 2026-04-21
Release C: collision-policy decision table + strict default + retire SKIP_STI_REHYDRATE (#1134)

### LINKED ISSUE #1134: refactor(core): decision-table register() + strict diagnostics default + retire rehydration flag

willgriffin:
Closes #1134. Third and final release of the [#1132 review](https://github.com/happyvertical/smrt/issues/1132) thread. Six commits, parity gates at commits 2 and 3.

## What changed

**Commit 1 — `collision-policy` module** ([5cb949e5](https://github.com/happyvertical/smrt/pull/new/refactor/issue-1134-collision-policy-strict-default/commits/5cb949e5))
New `packages/core/src/registry/collision-policy.ts` with a 16-row decision table + pure `decideCollisionPolicy(inputs)`. Every row names a scenario that maps 1:1 to an existing collision test (`issue-531`, `issue-555`, `issue-584`, `issue-847`, `issue-950`, `issue-951`, `issue-1000`). 23 unit tests cover every row plus ordering precedence.

**Commit 2 — Route `register()` through the table (parity gate)** ([a2c146b1](https://github.com/happyvertical/smrt/pull/new/refactor/issue-1134-collision-policy-strict-default/commits/a2c146b1))
The 11-branch exact-match + case-insensitive collision tree in `register()` (~300 LOC) collapses onto `applyRegisterCollisionPolicy(args)`. All state mutation stays in the caller; the policy function is pure. Full core suite + turbo 80/80 still green.

**Commit 3 — Route `registerFromManifest`, delete `resolveManifestCollision` (parity gate)** ([2c24b11c](https://github.com/happyvertical/smrt/pull/new/refactor/issue-1134-collision-policy-strict-default/commits/2c24b11c))
Same treatment for the manifest path. The `merge-manifest` caller contract is implemented end-to-end — `mergeManifestIntoExistingRegistration` followed by conditional key aliasing when the new qualified key differs from the canonical one. Preserves `issue-951:144-200` (qualified-key promotion during merge). `resolveManifestCollision` deleted.

Two decision-table refinements caught by the parity gate:
1. `manifest-stub-replacement` is now `origin === 'decorator'` only. Stub replacement never fired in `resolveManifestCollision` pre-C.
2. `manifest-same-package-merge` also fires when `existingHasNoPackage && hasNewQualifiedKey` — covers the issue-951 scenario where a bare source registration is promoted to a qualified key by an arriving manifest.

**Commit 4 — Retire `SMRT_SKIP_STI_REHYDRATE`** ([f7a11d10](https://github.com/happyvertical/smrt/pull/new/refactor/issue-1134-collision-policy-strict-default/commits/f7a11d10))
Flag deleted. Per Codex's review of the plan, the rehydration loop stays — the `external-runtime-hydration.test.ts:1071` stale-cache repair contract is preserved. Optimizing the loop away via eager invalidation is tracked in [#1139](https://github.com/happyvertical/smrt/issues/1139).

**Commit 5 — `SMRT_STRICT_REGISTRY` default-on** ([ede3cc34](https://github.com/happyvertical/smrt/pull/new/refactor/issue-1134-collision-policy-strict-default/commits/ede3cc34))
Severity-`'error'` diagnostics now throw at record time by default. Four codes affected: `MANIFEST_EXPORT_INVALID`, `MANIFEST_EXPORT_NOT_JSON`, `MANIFEST_EXPORT_NOT_FOUND`, `PACKAGE_MANIFEST_READ_FAILED`. Warn-severity codes unchanged. Opt-out: `SMRT_STRICT_REGISTRY=false`. One test (`register-package-manifest.test.ts::silently no-ops when the manifest is malformed JSON`) rewritten to expect a throw; a new opt-out test verifies `=false` restores the legacy behavior and keeps the diagnostic collected.

**Commit 6 — Changeset** ([3ec82c42](https://github.com/happyvertical/smrt/pull/new/refactor/issue-1134-collision-policy-strict-default/commits/3ec82c42))
Minor release on `@happyvertical/smrt-core`.

## Verification

- **Local**: `npx turbo build` 41/41 · `npx turbo test` 80/80 · smrt-core unit tests 1563/1563 pass (baseline 1539 + 24 new: 23 collision-policy rows + 1 strict-mode opt-out)
- **Parity**: every issue-specific collision test (#531, #555, #584, #847, #950, #951, #1000) passes unmodified

## Breaking changes

Minor. Two breaking flips:

| Change | Migration |
|---|---|
| `SMRT_STRICT_REGISTRY` default on | Fix the underlying manifest issue **or** set `SMRT_STRICT_REGISTRY=false` |
| `SMRT_SKIP_STI_REHYDRATE` removed | Remove the flag from your env; it's now a no-op |

Error message text is unchanged; the verbose-log output now includes a stable scenario id from the decision table.

## Review feedback addressed

All four points from the [Codex plan review](https://github.com/happyvertical/smrt/issues/1134#issuecomment-4290823035):

1. **§4 scope narrowed** — flag goes, rehydration loop stays; the `external-runtime-hydration.test.ts:1071` contract is preserved.
2. **`merge-manifest` caller contract documented** — the `applyManifestCollisionPolicy` branch in `class-registration.ts` calls `mergeManifestIntoExistingRegistration` then conditionally aliases the qualified key. `issue-951:144-200` verifies both steps.
3. **Invocation contract made explicit** — `decideCollisionPolicy`'s docstring calls out that ambiguous simple-name matches (multiple packages) must NOT be fed through the table. Caller-side handles the no-candidate and ambiguous-candidate cases.
4. **Diagnostic call-site count corrected** — 6 total (4 error, 2 warn), not 7. Test audit identified one test needing an opt-out stub; handled in commit 5.

## Test plan

- [x] Build all packages green (41/41)
- [x] Full test suite green (80/80 tasks, 1563 core tests)
- [x] Every row in the collision-policy table has a dedicated unit test
- [x] Every issue-specific collision test (531/555/584/847/950/951/1000) passes unmodified
- [x] `external-runtime-hydration.test.ts:1071` stale-cache repair test passes without the env flag
- [x] Strict-mode throws on malformed JSON; opt-out restores permissive behavior
- [ ] CI parity on PR

### DIFF
--- a/packages/core/src/registry/collision-policy.ts
+++ b/packages/core/src/registry/collision-policy.ts
@@ -113,9 +113,20 @@ export interface CollisionInputs {
 
   /**
    * `existing.packageName?.startsWith('@')` — existing entry came from a
-   * scoped package manifest (i.e., is externally identified).
+   * **scoped** package manifest (e.g. `@happyvertical/smrt-core`). Used by
+   * the external-package-replace (#584) and bundled-context scenarios
+   * that were gated on `@` in pre-Release-C code.
    */
   readonly existingIsPackageQualified: boolean;
+
+  /**
+   * Existing entry carries any packageName, scoped or not. Broader than
+   * `existingIsPackageQualified`. Used for the different-packages
+   * coexistence scenario where pre-Release-C code only required both
+   * packageNames to be present and different (unscoped manifests like
+   * `smrt-profiles` count here).
+   */
+  readonly existingHasAnyPackage: boolean;
 
   /**
    * Case-sensitive equality between `new.name` and `existing.name`.
@@ -243,19 +254,25 @@ const ROWS: readonly Row[] = [
   },
 
   {
+    // Uses `existingHasAnyPackage` not `existingIsPackageQualified`: the
+    // pre-Release-C `allowCoexistingQualifiedRegistration` check only
+    // required both packages to be present and differ, regardless of
+    // whether they were scoped (`@scope/pkg`) or unscoped (`smrt-profiles`).
+    // Narrowing to scoped would silently drop coexistence for unscoped
+    // manifests. See PR #1140 review (willgriffin P2 at line 251).
     scenario: 'manifest-different-packages-qualified-coexist',
     match: (i) =>
       i.origin === 'manifest' &&
       !i.samePackage &&
       i.hasNewQualifiedKey &&
-      i.existingIsPackageQualified,
+      i.existingHasAnyPackage,
     policy: 'coexist-qualified',
     reason: () =>
       'Different packages share a simple name; both registrations live under their qualified keys (issue #951).',
   },
 
   {
-    // Covers two related scenarios:
+    // Covers three related scenarios:
     //   (a) `samePackage`          — decorator + manifest from the same
     //       package land on the same class; merge fields (issue #1000).
     //   (b) `existingHasNoPackage` — source registration hasn't been
@@ -263,11 +280,19 @@ const ROWS: readonly Row[] = [
     //       the package identity (issue #951:144-200). The merge runs
     //       AND the caller aliases the existing entry under the new
     //       qualified key.
+    //   (c) The same scenarios when the existing entry is a manifest
+    //       stub from an earlier unqualified/local manifest registration
+    //       — the pre-Release-C code merged whenever
+    //       `!existing.packageName || packageName === existing.packageName`
+    //       regardless of stub status. Release C preserves that: removing
+    //       the `!existingIsManifestStub` gate lets a later
+    //       package-qualified manifest upgrade a stub's fields + alias
+    //       under the qualified key. See PR #1140 review (willgriffin P1
+    //       at line 270).
     scenario: 'manifest-same-package-merge',
     match: (i) =>
       i.origin === 'manifest' &&
       (i.samePackage || (i.existingHasNoPackage && i.hasNewQualifiedKey)) &&
-      !i.existingIsManifestStub &&
       i.hasManifestContent,
     policy: 'merge-manifest',
     reason: () =>
@@ -298,12 +323,15 @@ const ROWS: readonly Row[] = [
   },
 
   {
+    // pnpm-dup allows unscoped package names (old code didn't gate on
+    // scoped-only). `samePackage` already requires both sides to have a
+    // packageName and match, so no extra existingHasAnyPackage check
+    // is needed.
     scenario: 'decorator-case-insensitive-pnpm-duplicate',
     match: (i) =>
       i.origin === 'decorator' &&
       i.matchKind === 'case-insensitive' &&
       i.existingKeyIsQualified &&
-      i.existingIsPackageQualified &&
       i.samePackage &&
       i.sameName &&
       !i.sameConstructor,
--- a/packages/core/src/registry/collision-policy.ts
+++ b/packages/core/src/registry/collision-policy.ts
@@ -137,6 +137,31 @@ export interface CollisionInputs {
    * (old code required both).
    */
   readonly existingKeyIsQualified: boolean;
+
+  /**
+   * Manifest-origin only: the incoming `objectDef` carries at least one of
+   * `fields`, `methods`, `schema`, or `decoratorConfig`. Without that, the
+   * merge path has nothing to fold in and the caller should skip the
+   * merge-manifest policy entirely.
+   */
+  readonly hasManifestContent: boolean;
+
+  /**
+   * Manifest-origin only: the new registration's qualified key differs
+   * from the existing entry's canonical key. When true and the policy is
+   * `merge-manifest`, the caller MUST also alias the existing entry under
+   * the new qualified key (step 2 of the merge-manifest contract).
+   */
+  readonly registrationKeyDiffersFromExistingKey: boolean;
+
+  /**
+   * Existing registration has no `packageName` set. Distinct from
+   * `!samePackage`: this specifically means the existing entry is a bare
+   * source registration waiting for a package manifest to promote it. When
+   * true and the new registration carries a packageName, merge-manifest
+   * treats it as promotion rather than collision.
+   */
+  readonly existingHasNoPackage: boolean;
 }
 
 export interface PolicyDecision {
@@ -175,8 +200,15 @@ const ROWS: readonly Row[] = [
   },
 
   {
+    // Decorator-only: the decorator fires after a manifest builder
+    // pre-registered a stub placeholder under this name. The real class
+    // takes over. Manifest-origin stub-to-stub arrivals fall through to
+    // the source-file / STI rows instead.
     scenario: 'manifest-stub-replacement',
-    match: (i) => i.existingIsManifestStub && !i.sameConstructor,
+    match: (i) =>
+      i.origin === 'decorator' &&
+      i.existingIsManifestStub &&
+      !i.sameConstructor,
     policy: 'replace',
     reason: () =>
       'Existing entry is a manifest stub placeholder; the real implementation takes over.',
@@ -223,12 +255,23 @@ const ROWS: readonly Row[] = [
   },
 
   {
+    // Covers two related scenarios:
+    //   (a) `samePackage`          — decorator + manifest from the same
+    //       package land on the same class; merge fields (issue #1000).
+    //   (b) `existingHasNoPackage` — source registration hasn't been
+    //       promoted to a qualified key yet and a manifest now provides
+    //       the package identity (issue #951:144-200). The merge runs
+    //       AND the caller aliases the existing entry under the new
+    //       qualified key.
     scenario: 'manifest-same-package-merge',
     match: (i) =>
-      i.origin === 'manifest' && i.samePackage && !i.existingIsManifestStub,
+      i.origin === 'manifest' &&
+      (i.samePackage || (i.existingHasNoPackage && i.hasNewQualifiedKey)) &&
+      !i.existingIsManifestStub &&
+      i.hasManifestContent,
     policy: 'merge-manifest',
     reason: () =>
-      'Manifest arriving for an existing same-package source registration; fold fields in and alias the qualified key.',
+      'Manifest arriving for an existing same-package or unqualified source registration; fold fields in and alias the qualified key if keys differ.',
   },
 
   {
--- a/packages/core/src/registry/collision-policy.ts
+++ b/packages/core/src/registry/collision-policy.ts
@@ -1,0 +1,349 @@
+/**
+ * Release C (#1134) — unified collision policy for class registration.
+ *
+ * `register()` and `registerFromManifest()` both need to decide what to do
+ * when a new registration lands on an existing class with the same (or a
+ * case-insensitive-matching) name. Before Release C the decision logic was
+ * a ~300-line if/else tree duplicated across three code paths:
+ *
+ *   - `register()` exact-match branch          (class-registration.ts:159-275)
+ *   - `register()` case-insensitive branch     (class-registration.ts:286-391)
+ *   - `resolveManifestCollision()`             (class-registration.ts:876-929)
+ *
+ * This module collapses all three onto a single decision table. Each row
+ * names a scenario that exists in the current test suite; the row order
+ * encodes specificity (first match wins).
+ *
+ * # Invocation contract
+ *
+ * `decideCollisionPolicy(inputs)` is ONLY called when the caller has already
+ * identified exactly one colliding candidate for the new registration. Two
+ * cases MUST be handled upstream without consulting the table:
+ *
+ *   1. No colliding candidate exists  → register fresh.
+ *   2. Ambiguous simple name matches >1 existing entry across different
+ *      packages (`getCanonicalClassName()` returns undefined) → fall through
+ *      to the exact-key path and allow coexistence under qualified keys.
+ *      See issue #951 for the semantics being preserved.
+ *
+ * Feeding the ambiguous case through the table would lose the
+ * `coexist-qualified` behavior that's crucial for multi-package apps.
+ *
+ * @see https://github.com/happyvertical/smrt/issues/1134
+ */
+
+export type RegistrationOrigin = 'decorator' | 'manifest';
+
+/**
+ * How the colliding candidate was identified.
+ *
+ *   - `exact-key`          : `classes.get(name)` returned the candidate.
+ *                            `existingKey === name`.
+ *   - `case-insensitive`   : Iteration found a case-insensitive match on
+ *                            either the registry key or the entry's `name`
+ *                            property. `existingKey !== name`.
+ *   - `canonical-name`     : Looked up via `getCanonicalClassName(simpleName)`
+ *                            — unambiguous single match under a lowered
+ *                            simple name. Used by the manifest path.
+ */
+export type MatchKind = 'exact-key' | 'case-insensitive' | 'canonical-name';
+
+export type Policy =
+  /** Keep existing identity, update metadata via `upsertExistingEntry`. */
+  | 'accept'
+  /** Drop existing, register new as replacement. */
+  | 'replace'
+  /**
+   * Fold manifest fields into existing registration. The **caller MUST**:
+   *   (1) call `mergeManifestIntoExistingRegistration(existing, objectDef, packageName)`
+   *   (2) if `registrationKey !== existingCanonical`, alias the existing entry
+   *       under `registrationKey` via `classes.set()` + `constructorIndex.set()`
+   * Both steps together preserve the `issue-951:144-200` contract.
+   */
+  | 'merge-manifest'
+  /**
+   * Different packages, different qualified keys. Both registrations stay
+   * in the classes Map under their respective qualified keys. Ambiguity is
+   * surfaced at `findClass` time. Preserves issue #951.
+   */
+  | 'coexist-qualified'
+  /** Existing wins; silently ignore the new registration. */
+  | 'skip'
+  /** Unresolvable collision — raise a user-facing error. */
+  | 'throw';
+
+/**
+ * Every signal the decision table can read. Computed by the caller from
+ * inspection of the new and existing registrations.
+ */
+export interface CollisionInputs {
+  readonly origin: RegistrationOrigin;
+  readonly matchKind: MatchKind;
+
+  /** `new.ctor === existing.constructor` (decorator path only; false for manifest). */
+  readonly sameConstructor: boolean;
+
+  /** `new.sourceFile === existing.sourceFilePath`, both truthy. */
+  readonly sameSourceFile: boolean;
+
+  /** Both the new and existing registrations have known source files. */
+  readonly bothSourceFilesKnown: boolean;
+
+  /** `existing.constructor._isManifestStub === true`. */
+  readonly existingIsManifestStub: boolean;
+
+  /**
+   * New registration's source file lives in a bundled output directory
+   * (`.svelte-kit/output/`, `/dist/`, `/build/`, `.next/`, `.nuxt/`).
+   */
+  readonly newInBundledContext: boolean;
+
+  /**
+   * New class extends existing.
+   *   - decorator path: `new.ctor.prototype instanceof existing.constructor`
+   *   - manifest path : `new.objectDef.extends` names existing
+   */
+  readonly newExtendsExisting: boolean;
+
+  /** Inverse direction: existing extends new. */
+  readonly existingExtendsNew: boolean;
+
+  /** `new.packageName === existing.packageName` (both truthy). */
+  readonly samePackage: boolean;
+
+  /**
+   * `existing.packageName?.startsWith('@')` — existing entry came from a
+   * scoped package manifest (i.e., is externally identified).
+   */
+  readonly existingIsPackageQualified: boolean;
+
+  /**
+   * Case-sensitive equality between `new.name` and `existing.name`.
+   * Used to distinguish a pnpm-duplicate (same exact name) from a genuine
+   * case-insensitive collision (different names that lower to the same string).
+   */
+  readonly sameName: boolean;
+
+  /**
+   * New registration has (or can derive) a qualified key `@pkg:Name`.
+   * Paired with `existingIsPackageQualified` to drive the different-packages
+   * coexistence behavior from issue #951.
+   */
+  readonly hasNewQualifiedKey: boolean;
+
+  /**
+   * Existing registry key contains a `:` separator. Needed alongside
+   * `existingIsPackageQualified` to identify valid pnpm-dup candidates
+   * (old code required both).
+   */
+  readonly existingKeyIsQualified: boolean;
+}
+
+export interface PolicyDecision {
+  readonly policy: Policy;
+  /** Stable id. Matches the scenario name in the test suite 1-to-1. */
+  readonly scenario: string;
+  /** Human-readable explanation. Used by verboseLog + the error message. */
+  readonly reason: string;
+}
+
+interface Row {
+  readonly scenario: string;
+  readonly match: (inputs: CollisionInputs) => boolean;
+  readonly policy: Policy;
+  readonly reason: (inputs: CollisionInputs) => string;
+}
+
+/**
+ * Ordered decision table. First row whose `match` returns true wins.
+ * Order reflects specificity — scenarios that narrow down earlier
+ * scenarios come before them.
+ *
+ * Each row is named to match a scenario in the test suite so a reviewer
+ * can bisect one-to-one between "new class extends existing" (the code)
+ * and "sti-child-wins" (the test row and the verbose-log output).
+ */
+const ROWS: readonly Row[] = [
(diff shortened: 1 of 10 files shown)


packages/core/src/registry/collision-policy.ts
```diff
@@ -0,0 +1,392 @@
...
+      'Existing entry is the STI child of the incoming manifest parent; keep the child.',
+  },
+
+  {
+    scenario: 'manifest-different-packages-qualified-coexist',
+    match: (i) =>
+      i.origin === 'manifest' &&
+      !i.samePackage &&
+      i.hasNewQualifiedKey &&
+      i.existingIsPackageQualified,
```
willgriffin: [P2] This narrows cross-package coexistence compared with the old manifest path. Before this refactor, `allowCoexistingQualifiedRegistration` only required both package names to be present and different; it did not require the existing package to be scoped. `createQualifiedName()` still happily builds `packageName:ClassName` for unscoped package ids, and this repo already registers manifests with unscoped names like `smrt-profiles`. With the new `existingIsPackageQualified` gate, a second unscoped package that defines the same simple class name now falls through to `manifest-default-skip`, silently dropping the registration instead of coexisting.

packages/core/src/registry/collision-policy.ts
```diff
@@ -0,0 +1,392 @@
...
+    //   (b) `existingHasNoPackage` — source registration hasn't been
+    //       promoted to a qualified key yet and a manifest now provides
+    //       the package identity (issue #951:144-200). The merge runs
+    //       AND the caller aliases the existing entry under the new
+    //       qualified key.
+    scenario: 'manifest-same-package-merge',
+    match: (i) =>
+      i.origin === 'manifest' &&
+      (i.samePackage || (i.existingHasNoPackage && i.hasNewQualifiedKey)) &&
+      !i.existingIsManifestStub &&
```
willgriffin: [P1] The `existingHasNoPackage` promotion path no longer covers manifest-first registrations. `registerFromManifest()` always creates a manifest stub, so when an unqualified/local manifest seeds `ClassName` first and a later package-qualified manifest for that same class arrives, this row does not match because of `!i.existingIsManifestStub`. The flow falls through to `manifest-default-skip`, so the later package identity and any richer schema/validationRules never merge or alias under `@pkg:ClassName`. Pre-Release-C, the corresponding branch merged whenever `!existing.packageName || packageName === existing.packageName`, regardless of stub status.

willgriffin:
All three inline comments plus a deep-review pass from two subagents addressed in [1aa44cc0](https://github.com/happyvertical/smrt/pull/1140/commits/1aa44cc0).

### Correctness fixes

**P1 Copilot — `hasNewQualifiedKey` derivation** ([class-registration.ts:287](packages/core/src/registry/class-registration.ts:287))
Switched to `isQualifiedName(registrationKey)`. Now covers the `objectDef.qualifiedName` fallback path that bypasses `packageName`.

**P2 willgriffin — `existingIsPackageQualified` too narrow** ([collision-policy.ts:251](packages/core/src/registry/collision-policy.ts:251))
Added `existingHasAnyPackage: boolean` to `CollisionInputs`. The `manifest-different-packages-qualified-coexist` row and the `decorator-case-insensitive-pnpm-duplicate` row both use the broader semantic. Unscoped package names (`smrt-profiles`, etc.) now coexist instead of falling through to skip. Dedicated unit test added at `collision-policy.test.ts`.

**P1 willgriffin — stub-upgrade blocked by `!existingIsManifestStub`** ([collision-policy.ts:270](packages/core/src/registry/collision-policy.ts:270))
Removed the gate. The `manifest-same-package-merge` row now fires for stub-upgrade cases too, matching pre-Release-C behavior where merge ran whenever `packageName === existing.packageName || !existing.packageName`. Dedicated unit test added.

### Additional findings from a parallel subagent review

**Correctness — `existingExtendsNew` missing `registrationKey` comparison** ([class-registration.ts:260](packages/core/src/registry/class-registration.ts:260))
Symmetric to `newExtendsExisting`, which already includes the check. Caught a case where a child's `existing.extends` stores a qualified parent key post-`qualifyExtendsName()` and the parent manifest then arrives again — pre-fix would register a duplicate parent instead of skipping. New regression test at `issue-950-manifest-sti-collision.test.ts`.

**Migration — thrown error message opt-out hint** ([diagnostics.ts:87](packages/core/src/registry/diagnostics.ts:87))
Appended ` — set SMRT_STRICT_REGISTRY=false to record this diagnostic without throwing.` The hint is load-bearing because `PACKAGE_MANIFEST_READ_FAILED` can propagate out of a `__smrt-register__.ts` top-level side effect at module import time — before any consumer code has a chance to run. Without the hint, the crash has no visible recovery path.

**Migration — deprecation warn for `SMRT_SKIP_STI_REHYDRATE`** ([object.ts:83](packages/core/src/object.ts:83))
One-shot warn fires the first time an STI `save()` runs with the env var set. Anyone who followed the Release A changeset guidance and set the flag now discovers it's been retired instead of silently losing the behavior they wanted.

**Coverage — `MANIFEST_EXPORT_*` codes had no tests** (new file `strict-registry-manifest-export-errors.test.ts`)
Each of the three codes now has both a throw-path test and an `SMRT_STRICT_REGISTRY=false` opt-out-path test, plus a dedicated check that the thrown error message includes the opt-out hint. 7 tests total.

### Verification

- 1573/1573 core tests pass (baseline 1563 + 10 new)
- `npx turbo test` — 80/80 tasks green
- Every decision-table row still has a named scenario test; the three new scenarios (unscoped coexist, stub-upgrade merge, qualified-extends STI parent-skip) each have explicit coverage

PR #1167, opened 2026-04-29
feat(analytics): add Matomo provider

willgriffin:
## Summary
- add `AnalyticsProvider.MATOMO` to `@happyvertical/smrt-analytics`
- add Matomo model and collection helpers
- update docs, package metadata, Svelte provider label, manifest, and tests
- update SMRT SDK package catalog/overrides/lockfile to the released `@happyvertical/*` SDK `0.72.0` train

## Release train context
This gives Anytown a first-class SMRT analytics provider value for centrally managed Matomo site config, now aligned with the released SDK access-verification train.

## Verification
- `pnpm --filter @happyvertical/smrt-analytics exec vitest run src/__tests__/analytics-property.test.ts`
- `pnpm --filter @happyvertical/smrt-analytics check`
- `pnpm --filter @happyvertical/smrt-analytics --filter @happyvertical/smrt-agents --filter @happyvertical/smrt-images --filter @happyvertical/smrt-jobs build`
- `bash scripts/check-sdk-versions.sh`
- `git diff --check`
- pre-push format and workflow validation hook passed


packages/analytics/src/manifest/manifest.json
```diff
@@ -1104,6 +1104,14 @@
           "isStatic": false,
           "isPublic": true
         },
+        "isMatomo": {
```
willgriffin, line 1107: The committed src manifest was hand-edited rather than regenerated — the timestamp field (`1767322692161`) is unchanged from main, but `ManifestBuilder.generate()` in packages/vitest/src/index.ts always stamps a fresh value. Mostly harmless because `smrtVitestPlugin()` regenerates `.smrt/manifest.json` at test startup, but `src/manifest/manifest.json` is also a documented fallback path (packages/core/src/manifest/manager.ts:28, packages/cli/src/discovery/manifest-discovery.ts:109). A downstream consumer hitting that fallback would load a manifest that drifts from what the scanner would actually produce. Recommend regenerating with `smrt generate:test` instead of hand-patching.

packages/analytics/src/models/AnalyticsProperty.ts
```diff
@@ -58,7 +60,7 @@ export class AnalyticsProperty extends SmrtObject {
   apiSecret: string = '';
 
   /**
-   * Site domain for Plausible
+   * Site domain for Plausible/Matomo
```
willgriffin, line 63: Adjacent observation: the `apiSecret` field a few lines above (line 60) still has its docstring as "API secret for server-side tracking (GA4)". Matomo uses `token_auth` for Reporting API access — after this PR a developer adding a Matomo site has no obvious place to store the auth token. Consider updating the `apiSecret` docstring to note that Matomo's `token_auth` is stored here, adding an explicit field, or documenting that Matomo credentials live in `providerMetadata` JSON. As-is this will become friction the moment anyone tries to wire up actual Matomo API calls.

packages/analytics/src/collections/AnalyticsPropertyCollection.ts
```diff
@@ -43,9 +43,9 @@ export class AnalyticsPropertyCollection extends SmrtCollection<AnalyticsPropert
   }
 
   /**
-   * Find property by site domain (Plausible)
+   * Find property by site domain (Plausible/Matomo)
```
willgriffin: Doc says `(Plausible/Matomo)` but `findBySiteDomain` only filters on `siteDomain` — not on provider. If two providers ever share a domain string (during migration or testing) the query returns whichever the DB picked first. Not a bug per se, but the comment now implies a stronger guarantee than the query offers. Either tighten the query (`where: { siteDomain, provider: { in: [PLAUSIBLE, MATOMO] } }`) or soften the doc.

packages/analytics/src/svelte/PropertyInfo.svelte
```diff
@@ -24,7 +24,9 @@ const providerLabel = $derived(
     ? 'Google Analytics 4'
     : provider === 'plausible'
       ? 'Plausible'
-      : (provider ?? 'Unknown'),
+      : provider === 'matomo'
+        ? 'Matomo'
+        : (provider ?? 'Unknown'),
```
willgriffin: Style nit: this 4-branch nested ternary is starting to bend. A small lookup map would be cleaner and trivially extensible:

```ts
const PROVIDER_LABELS: Record<string, string> = {
  ga4: 'Google Analytics 4',
  plausible: 'Plausible',
  matomo: 'Matomo',
};
const providerLabel = $derived(
  (provider && PROVIDER_LABELS[provider]) ?? provider ?? 'Unknown',
);
```

Not a blocker.

packages/analytics/src/__tests__/analytics-property.test.ts
```diff
@@ -177,13 +204,24 @@ describe.skipIf(skipTests)(
...
       expect(ga4Properties).toHaveLength(1);
       expect(ga4Properties[0].provider).toBe(AnalyticsProvider.GA4);
 
       const plausibleSites = await collection.findPlausibleSites();
       expect(plausibleSites).toHaveLength(1);
       expect(plausibleSites[0].provider).toBe(AnalyticsProvider.PLAUSIBLE);
+
+      const matomoSites = await collection.findMatomoSites();
+      expect(matomoSites).toHaveLength(1);
+      expect(matomoSites[0].provider).toBe(AnalyticsProvider.MATOMO);
```
willgriffin, line 250: Test gap: the new `findMatomoSites` is covered, but `findBySiteDomain` (whose doc was just expanded to claim Matomo applicability) has no Matomo-side test. A one-liner — create a Matomo property with a siteDomain, then `findBySiteDomain` it back — would close the loop with the doc change.

willgriffin (commented):
Deep review of the Matomo provider addition. Overall the change is small, symmetric with existing GA4/Plausible plumbing, and well-scoped to provider config (no SDK/runtime integration claimed in the description, so I didn't grade against that).

Left line comments on five things — none are blockers, but two are worth resolving before merge:

1. **Manifest looks hand-edited.** `src/manifest/manifest.json` got the new method entries but its `timestamp` field is unchanged from main, while `ManifestBuilder.generate()` always stamps fresh. Recommend regenerating the manifest properly so the checked-in fallback doesn't drift from scanner output.
2. **`apiSecret` doc is now incomplete for Matomo.** Matomo's `token_auth` has no documented home. Pick one of: update the docstring, add a field, or document `providerMetadata` as the storage location.
3. `findBySiteDomain` doc claims Plausible/Matomo applicability but doesn't filter on provider — minor doc-vs-impl gap.
4. `PropertyInfo.svelte` ternary chain is getting cluttered; lookup map suggested.
5. `findBySiteDomain` has no Matomo-side test.

Out of scope, just to set expectations: no Matomo SDK integration, no `AnalyticsEvent` Matomo branch, no DataStream model for Matomo. The PR description scopes this to "centrally managed Matomo site config", so that all reads as intentional.

Approve with the doc/manifest nits.

PR #1172, opened 2026-04-30
[codex] Handle legacy STI discriminator saves

### LINKED ISSUE #1170: Handle legacy STI discriminator duplicates without primary-key collisions
## Summary

Anytown production had legacy `places` rows with `_meta_type = Place` alongside qualified `@happyvertical/smrt-places:Place` rows for the same `slug` + `context`. When Caelus loaded a legacy row and saved it, serialization upgraded the discriminator to the qualified type while preserving the legacy id. The generated upsert then collided with the existing qualified row / primary key path instead of reconciling the legacy row safely.

## Impact

This caused scheduled Caelus weather fetch jobs to fail with duplicate-key errors until the live data was repaired. The application-level repair merged legacy duplicate Place rows into the qualified rows, but SMRT should ideally detect or reconcile this STI migration shape upstream.

## Expected behavior

When a hydrated legacy STI row serializes to a qualified discriminator, save/migration/reconcile logic should avoid creating a second identity collision. Possible approaches:

- migrate legacy discriminators before normal saves run
- detect same logical identity with different discriminator and merge/retarget safely
- fail with a targeted actionable migration error instead of a generic duplicate-key upsert failure

## Notes

This is not a request for app-level workarounds. The app repair is repeatable, but the underlying STI discriminator transition should be handled by SMRT so other consumers do not trip over the same shape.

willgriffin:
## Summary

- Resolves #1170 by detecting legacy STI discriminator upgrades before the normal `(slug, context, _meta_type)` upsert runs.
- Updates a lone legacy row in place by primary key so `Place` -> `@happyvertical/smrt-places:Place` migrations do not try to insert a second row with the same id.
- Raises `DB_STI_DISCRIMINATOR_CONFLICT` with row ids and logical identity when both legacy and qualified rows already exist.
- Adds regression coverage for both the safe in-place upgrade and the duplicate-row conflict path.

## Review Notes

During self-review I found the first pass did discriminator planning outside the save retry path. I tightened that by retrying the new read probes while still keeping the intentional duplicate conflict non-retryable.

## Validation

- `pnpm --filter @happyvertical/smrt-core exec vitest run src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts src/__tests__/sti-meta-fields.test.ts src/__tests__/sti-polymorphic-queries.test.ts src/__tests__/issue-1165-tenant-upsert-repair.test.ts`
- `pnpm --filter @happyvertical/smrt-core typecheck`
- `pnpm exec biome check packages/core/src/object.ts packages/core/src/errors.ts packages/core/src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts`
- `pnpm --filter @happyvertical/smrt-core test`
- `pnpm --filter @happyvertical/smrt-core build`
- pre-push `format-check` and workflow validation

### DIFF
--- a/packages/cli/src/commands/sti-upgrade.ts
+++ b/packages/cli/src/commands/sti-upgrade.ts
@@ -4,6 +4,10 @@ import {
   ObjectRegistry,
 } from '@happyvertical/smrt-core';
 
+type QueryableDb = {
+  query(sql: string, ...params: any[]): Promise<any>;
+};
+
 export type StiDiscriminatorUpgradeResult =
   | {
       action: 'skip';
@@ -15,6 +19,25 @@ export type StiDiscriminatorUpgradeResult =
       currentQualifiedName: string;
       sourceKind: 'simple' | 'stale-qualified';
     };
+
+export interface StiDiscriminatorRepairConflict {
+  tableName: string;
+  legacyMetaType: string;
+  qualifiedMetaType: string;
+  legacyId: string | null;
+  qualifiedId: string | null;
+  conflictIdentity: Record<string, any>;
+}
+
+export interface StiDiscriminatorRepairResult {
+  tableName: string;
+  legacyMetaType: string;
+  qualifiedMetaType: string;
+  checkedRows: number;
+  updatedRows: number;
+  wouldUpdateRows: number;
+  conflicts: StiDiscriminatorRepairConflict[];
+}
 
 export function resolveStiDiscriminatorUpgrade(
   metaType: string,
@@ -49,4 +72,122 @@ export function resolveStiDiscriminatorUpgrade(
     currentQualifiedName,
     sourceKind: isQualifiedName(metaType) ? 'stale-qualified' : 'simple',
   };
+}
+
+function quoteIdentifier(name: string): string {
+  return `"${name.replace(/"/g, '""')}"`;
+}
+
+function rowsFromResult(result: any): any[] {
+  if (Array.isArray(result)) {
+    return result;
+  }
+  return Array.isArray(result?.rows) ? result.rows : [];
+}
+
+function getRowCount(result: any): number | undefined {
+  return typeof result?.rowCount === 'number' ? result.rowCount : undefined;
+}
+
+function buildNullSafeIdentityPredicate(columns: string[]): string {
+  return columns
+    .map((column) => {
+      const quoted = quoteIdentifier(column);
+      return `(${quoted} = ? OR (${quoted} IS NULL AND ? IS NULL))`;
+    })
+    .join(' AND ');
+}
+
+function getIdentityParams(row: Record<string, any>, columns: string[]): any[] {
+  const params: any[] = [];
+  for (const column of columns) {
+    params.push(row[column], row[column]);
+  }
+  return params;
+}
+
+export function getStiConflictIdentityColumns(className: string): string[] {
+  return ObjectRegistry.getConflictColumns(className).filter(
+    (column) => column !== '_meta_type',
+  );
+}
+
+export async function repairStiDiscriminatorRows(options: {
+  db: QueryableDb;
+  tableName: string;
+  className: string;
+  legacyMetaType: string;
+  qualifiedMetaType: string;
+  dryRun?: boolean;
+}): Promise<StiDiscriminatorRepairResult> {
+  const {
+    db,
+    tableName,
+    className,
+    legacyMetaType,
+    qualifiedMetaType,
+    dryRun = false,
+  } = options;
+  const identityColumns = getStiConflictIdentityColumns(className);
+  const selectColumns = Array.from(new Set(['id', ...identityColumns]));
+  const quotedTable = quoteIdentifier(tableName);
+  const selectSql = `SELECT ${selectColumns.map(quoteIdentifier).join(', ')} FROM ${quotedTable} WHERE ${quoteIdentifier('_meta_type')} = ?`;
+  const legacyRows = rowsFromResult(await db.query(selectSql, legacyMetaType));
+  const result: StiDiscriminatorRepairResult = {
+    tableName,
+    legacyMetaType,
+    qualifiedMetaType,
+    checkedRows: legacyRows.length,
+    updatedRows: 0,
+    wouldUpdateRows: 0,
+    conflicts: [],
+  };
+
+  for (const row of legacyRows) {
+    const conflictIdentity = Object.fromEntries(
+      identityColumns.map((column) => [column, row[column]]),
+    );
+    const identityPredicate = buildNullSafeIdentityPredicate(identityColumns);
+    const duplicateSql = [
+      `SELECT ${quoteIdentifier('id')} FROM ${quotedTable}`,
+      `WHERE ${quoteIdentifier('_meta_type')} = ?`,
+      identityPredicate ? `AND ${identityPredicate}` : '',
+      'LIMIT 1',
+    ]
+      .filter(Boolean)
+      .join(' ');
+    const duplicateRows = rowsFromResult(
+      await db.query(
+        duplicateSql,
+        qualifiedMetaType,
+        ...getIdentityParams(row, identityColumns),
+      ),
+    );
+
+    if (duplicateRows.length > 0) {
+      result.conflicts.push({
+        tableName,
+        legacyMetaType,
+        qualifiedMetaType,
+        legacyId: row.id ?? null,
+        qualifiedId: duplicateRows[0].id ?? null,
+        conflictIdentity,
+      });
+      continue;
+    }
+
+    if (dryRun) {
+      result.wouldUpdateRows++;
+      continue;
+    }
+
+    const updateResult = await db.query(
+      `UPDATE ${quotedTable} SET ${quoteIdentifier('_meta_type')} = ? WHERE ${quoteIdentifier('id')} = ?`,
+      qualifiedMetaType,
+      row.id,
+    );
+    result.updatedRows += getRowCount(updateResult) ?? 1;
+  }
+
+  return result;
 }
--- a/packages/core/src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts
+++ b/packages/core/src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts
@@ -4,10 +4,18 @@
  */
 
 import type { DatabaseInterface } from '@happyvertical/sql';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  beforeEach,
+  describe,
+  expect,
+  it,
+} from 'vitest';
 import { field } from '../decorators';
 import { SmrtObject } from '../object';
-import { smrt } from '../registry';
+import { ObjectRegistry, smrt } from '../registry';
 import { getTestDatabase } from '../testing/database';
 
 @smrt({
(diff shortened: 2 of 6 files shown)


packages/core/src/object.ts
```diff
@@ -271,6 +275,111 @@ export class SmrtObject extends SmrtClass {
...
+    if (existingQualified && existingQualified.id !== id) {
+      throw DatabaseError.stiDiscriminatorConflict({
+        className,
+        tableName: this.tableName,
+        id,
+        slug,
+        context,
+        legacyMetaType: currentMetaType,
+        qualifiedMetaType,
+        duplicateId: String(existingQualified.id),
```
willgriffin, line 385: **In-memory `_meta_type` is stale after a successful `updateById`.**

After the update, the row in the DB is `qualifiedMetaType`, but `this._meta_type` still holds the legacy value. The next `save()` on the same instance will re-run `planPersistenceWrite` and pay for two more probe reads before hitting the `existingById._meta_type !== currentMetaType` early-return. Functionally correct, but wasteful — and it's the kind of thing that quietly stays slow in production loops that hold instances across saves.

Suggest threading `qualifiedMetaType` out of the planner (e.g. include it on the `updateById` plan variant) and calling `this.setMetaType(qualifiedMetaType)` after the update succeeds. One-liner, pays back on every subsequent save.

packages/core/src/object.ts
```diff
@@ -271,6 +275,111 @@ export class SmrtObject extends SmrtClass {
...
+      async () => {
+        try {
+          return await this.db.get(this.tableName, {
+            slug,
+            context,
+            _meta_type: qualifiedMetaType,
+          });
+        } catch (error) {
+          throw DatabaseError.queryFailed(
+            `get(${this.tableName}, { slug: ${slug}, context: ${context}, _meta_type: ${qualifiedMetaType} })`,
```
willgriffin: **Hardcoded `slug`/`context` for the duplicate probe ignores custom `conflictColumns`.**

The planner accepts `conflictColumns` as a parameter but then probes the qualified duplicate by `{ slug, context, _meta_type }` regardless. Today no STI class in the monorepo overrides `conflictColumns` (I grepped — STI + custom conflict cols is empty), so this is fine in practice, but the unused parameter is misleading and the assumption isn't documented.

Two reasonable options:
1. Use `conflictColumns.filter(c => c !== '_meta_type')` to build the probe filter, so the logic stays correct if someone later adds a custom-keyed STI class.
2. Drop the unused `conflictColumns` parameter and add an early-return-with-comment when `conflictColumns` differs from `['slug', 'context', '_meta_type']`, explicitly scoping this fix to the default STI key shape.

packages/core/src/object.ts
```diff
@@ -271,6 +275,111 @@ export class SmrtObject extends SmrtClass {
...
+          );
+        }
+      },
+      3,
+      500,
+    );
+
+    if (existingQualified && existingQualified.id !== id) {
+      throw DatabaseError.stiDiscriminatorConflict({
+        className,
```
willgriffin, line 377: **Non-atomic plan + write — worth a comment, probably out of scope to fix.**

`existingById`, `existingQualified`, and the eventual `UPDATE WHERE id = ?` are three independent statements. Two processes that both load the same legacy row can both pass the `existingQualified === null` check and both issue updateById against the same id — last-writer-wins, which matches existing save semantics, but is worth a one-line comment near the planning block so future readers know this isn't transactional. A real fix would need a tx wrapping read + update.

packages/core/src/object.ts
```diff
@@ -1230,10 +1354,11 @@ export class SmrtObject extends SmrtClass {
                 const field = this.extractConstraintField(error.message);
                 throw ValidationError.requiredField(field, className);
               }
-              throw DatabaseError.queryFailed(
-                `UPSERT INTO ${this.tableName}`,
-                error,
-              );
+              const operation =
+                writePlan.type === 'updateById'
+                  ? `UPDATE ${this.tableName} WHERE id = ${data.id}`
```
willgriffin: **Operation label interpolates the raw id.**

`UPDATE ${this.tableName} WHERE id = ${data.id}` puts the user-supplied id into a user-facing error string. Not a SQL injection concern (this is just an error label, not executed SQL), but the existing UPSERT label doesn't include the id, so it's a small inconsistency. Consider `UPDATE ${this.tableName} (id-targeted)` or just `UPDATE ${this.tableName}` to match the existing style.

packages/core/src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+@smrt({
+  tableStrategy: 'sti',
+  tableName: 'issue_1170_places',
+})
+class Issue1170Place extends SmrtObject {
+  @field()
+  name: string = '';
+}
+
```
willgriffin, line 29: **Test class lives in core, so the qualified name is `@happyvertical/smrt-core:Issue1170Place`.**

The real bug from #1170 is about a `@happyvertical/smrt-places:Place` row written before the qualified-name migration. The fix should still work for cross-package classes (the logic only depends on `currentMetaType.includes(':')` and `isValidMetaType`), but the regression test doesn't exercise that shape. A second test that registers the test class as if it lived in a different package — or asserts the qualified name comes from a non-core namespace — would more directly mirror the production failure.

willgriffin (commented):
Deep review — overall LGTM with one minor blocker and a couple of follow-ups. The fix is tightly scoped (the slow path only fires on legacy → qualified upgrades), the error is actionable with structured details, and tests cover both the in-place upgrade and the duplicate-conflict path. Comments inline.

packages/cli/src/commands/sti-upgrade.ts
```diff
@@ -50,3 +73,121 @@ export function resolveStiDiscriminatorUpgrade(
...
+  className: string;
+  legacyMetaType: string;
+  qualifiedMetaType: string;
+  dryRun?: boolean;
+}): Promise<StiDiscriminatorRepairResult> {
+  const {
+    db,
+    tableName,
+    className,
+    legacyMetaType,
```
willgriffin, line 127: **Per-row loop replaces a single bulk UPDATE — material regression for large migrations.**

The previous CLI repair path was one statement: `UPDATE table SET _meta_type=? WHERE _meta_type=?`. The new path does:

1. `SELECT id, ...identity FROM table WHERE _meta_type = legacy` — 1 query
2. For each legacy row: `SELECT id FROM table WHERE _meta_type = qualified AND <identity predicate>` — N queries
3. For each non-conflicting row: `UPDATE ... WHERE id = ?` — up to N more queries

For a clean migration with no conflicts on a 1M-row table that's ~2M round-trips instead of 1. Worth a fast path: do one `SELECT 1 FROM table t1 WHERE _meta_type=legacy AND EXISTS (SELECT 1 FROM table t2 WHERE t2._meta_type=qualified AND <identity match>) LIMIT 1` first; if no conflicts, fall through to the bulk UPDATE; only enter the per-row probe-and-update when conflicts exist.

Alternative if SQLite-only is too restrictive: a single `INSERT INTO ... SELECT ... ON CONFLICT DO NOTHING` style merge. But the EXISTS-fast-path is the simplest preserve-correctness optimization.

packages/core/src/object.ts
```diff
@@ -271,6 +275,122 @@ export class SmrtObject extends SmrtClass {
...
+    ) {
+      return upsertPlan;
+    }
+    const qualifiedMetaType = String(nextMetaType);
+    const conflictIdentity: Record<string, any> = {};
+    for (const column of conflictColumns.filter(
+      (conflictColumn) => conflictColumn !== '_meta_type',
+    )) {
+      if (!Object.hasOwn(data, column)) {
+        return upsertPlan;
```
willgriffin, line 330: **Silent fallback when a custom conflict column isn't in `data`.**

```ts
if (!Object.hasOwn(data, column)) {
  return upsertPlan;
}
```

If an STI class declares a custom conflict column that isn't a top-level field on the row (e.g. it's a `@meta()` field nested into `_meta_data`, or a getter that toJSON skips), we silently fall back to upsert — which will then fail with the original primary-key collision that this PR is meant to prevent. Operators end up back in the broken state with no signal that the planner bailed.

Two options:
- Throw a targeted error here (e.g. `DatabaseError.stiDiscriminatorConflict` variant or a new code) explaining that the conflict column wasn't in serialized output, so the legacy-upgrade path can't run.
- At minimum, `console.warn` once with `className` + missing column name so it's not invisible.

Materialistically this might never fire today, but the silent-bail-and-fail combination is exactly the shape that produced #1170 in the first place.

packages/core/src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts
```diff
@@ -0,0 +1,263 @@
...
+const originalExternalQualifiedName = externalPlaceClass?.qualifiedName;
+
+describe('Issue #1170: legacy STI discriminator save handling', () => {
+  let db: DatabaseInterface;
+
+  beforeAll(() => {
+    if (!externalPlaceClass) {
+      throw new Error('Issue1170ExternalPlace was not registered');
+    }
+    externalPlaceClass.packageName = '@happyvertical/smrt-places';
```
willgriffin, line 75: **Direct mutation of `externalPlaceClass.packageName` / `.qualifiedName` is a test backdoor.**

```ts
externalPlaceClass.packageName = '@happyvertical/smrt-places';
externalPlaceClass.qualifiedName = externalQualifiedMetaType;
```

This works because the registered class entry is a mutable object and there's no API stopping you, but it bypasses every guarantee the registry would normally enforce on registration. If the registry ever adds derived state from `packageName` (e.g. caches qualifiedName at register time, or indexes classes by package), this test will silently diverge from real behavior.

Not a blocker — the alternative (a real second-package fixture) is a lot more setup. But please add a comment above the mutation explaining *why* this shortcut is acceptable here, and consider exposing a test-only registry helper (`ObjectRegistry.__overrideForTest({ packageName, qualifiedName })`) so future tests don't reach into the private shape.

willgriffin (commented):
Follow-up review on the three new commits. The four blocking items from my earlier pass are all addressed:

- ✅ In-memory `_meta_type` is now refreshed via `setMetaType(writePlan.qualifiedMetaType)` after `updateById` succeeds.
- ✅ Custom `conflictColumns` are honored: `planPersistenceWrite` now builds `conflictIdentity` from `conflictColumns.filter(c => c !== '_meta_type')`, and there's a regression test (`Issue1170CustomIdentityPlace` with `['external_key', '_meta_type']`).
- ✅ Race comment added at the `updateById` return.
- ✅ Operation label collapsed to `UPDATE ${tableName} (id-targeted)` and the probe `queryFailed` labels no longer interpolate raw values.
- ✅ Cross-package test shape now exists via `Issue1170ExternalPlace` whose qualified name is mutated to `@happyvertical/smrt-places:Issue1170ExternalPlace`.

Bonus scope: `repairStiDiscriminatorRows` was added to `smrt db:migrate --upgrade-sti` so operators can detect/fix the same shape proactively. That's a nice extension.

A couple of new things I'd raise — see inline.

packages/core/src/object.ts
```diff
@@ -1211,11 +1331,27 @@ export class SmrtObject extends SmrtClass {
...
+        conflictColumns,
+      );
 
       await ErrorUtils.withRetry(
         async () => {
           try {
-            await this.db.upsert(this.tableName, conflictColumns, data);
+            if (writePlan.type === 'updateById') {
+              const { id: _id, ...updateData } = data;
+              await this.db.update(this.tableName, { id: data.id }, updateData);
```
willgriffin, line 1346: Nice — `setMetaType(writePlan.qualifiedMetaType)` here closes the stale-instance path I flagged in the previous review. One small nit: this side effect happens inside the `withRetry` callback, so on a transient `db.update` failure that succeeds on retry, the meta type is set after the eventual success, which is correct — but if `db.update` partially succeeds and then throws (write-but-error), the in-memory type still reflects the legacy value. That matches existing save semantics (no rollback of in-memory state on partial write), so it's fine, just worth flagging in case it ever surfaces.

PR #1182, opened 2026-05-01
feat(media): add media bundle persistence helpers

willgriffin:
## Summary
- Add structural media-bundle persistence helpers for @happyvertical/smrt-video and @happyvertical/smrt-images.
- Persist primary assets, retained support-file assets/associations, metadata artifacts, and GPS tracks through explicit adapters.
- Cover hidden-retained defaults, drop-after-extract sidecars, and missing GPS adapter warnings with tests.

## Validation
- pnpm --filter @happyvertical/smrt-video test -- media-bundle-persistence.test.ts
- pnpm --filter @happyvertical/smrt-images test -- media-bundle-persistence.test.ts
- pnpm --filter @happyvertical/smrt-video build
- pnpm --filter @happyvertical/smrt-images build

Note: the video package build exits 0 while emitting existing vite:dts diagnostics about unbuilt workspace package declarations in this clean worktree.


willgriffin (commented):
## Review — feat(media): media bundle persistence helpers

### Overview
Adds `persistMediaBundleInspection` (video) and `persistImageMediaBundleInspection` (images), each with a structural inspection type and an adapter interface for `upsertAsset` / `associateSupportFile` / `writeMetadataArtifact` / `replaceGpsTrack`. Behavior: writes primary, retains/skips support files by visibility, optionally emits a metadata artifact and a primary GPS track. Test coverage includes the hidden-retained happy path and the drop-after-extract + missing GPS adapter path.

### High priority

**Near-total duplication between the two new modules.** [packages/video/src/media-bundle-persistence.ts](packages/video/src/media-bundle-persistence.ts) and [packages/images/src/media-bundle-persistence.ts](packages/images/src/media-bundle-persistence.ts) are ~180 lines each and structurally identical — the only differences are the `MediaBundle*` vs `ImageMediaBundle*` type prefix and two trivial style divergences (see below). Tests are similarly mirrored. Any future bug fix or contract change must be made in two places, and consumers must pick between two indistinguishable APIs depending on which package they imported from. Consider hoisting the helper + types into a shared home (e.g. `@happyvertical/smrt-assets`, which already owns provider-agnostic asset management and `AssetAssociation`) and re-exporting from `images`/`video` if needed for ergonomics. CLAUDE.md is fine with thin packages of three similar lines, but ~180 mirrored lines across packages is the case the rule was written to prevent.

### Medium priority

- **Stylistic drift between the two copies.** images uses optional chaining `adapter.associateSupportFile?.(...)` and a ternary + `null` for `metadataArtifactId`; video uses an `if (adapter.associateSupportFile)` guard and `let metadataArtifactId: string | undefined`. Same effective behavior, but worth converging when consolidated.
- **Support-file GPS tracks are silently ignored.** `replaceGpsTrack` is only invoked for `inspection.metadata.gpsTrack`. The video fixture in [packages/video/src/media-bundle-persistence.test.ts:32](packages/video/src/media-bundle-persistence.test.ts:32) includes a `gpsTrack` on a support file's metadata, but that track is never passed to the adapter and the test does not assert that it should be. If proxy/sidecar GPS data is intentionally non-canonical, a one-liner comment on the inspection loop would prevent a future contributor from \"fixing\" the apparent oversight.
- **`gpsTrackPointCount` reports input length, not persisted length.** The adapter is `Promise<void>`, so the helper has no way to know how many points were actually written (e.g. after dedup). Either rename to `gpsTrackPointInputCount` or have `replaceGpsTrack` return the persisted count.

### Low priority

- **`*InspectionLike` naming.** The `Like` suffix conventionally signals a structural dual of a richer named interface. If no fuller `MediaBundleInspection` exists yet, the suffix is misleading — drop it or document the intent.
- **Empty line removed from [packages/video/src/index.ts](packages/video/src/index.ts) before `VideoCompositionAsset`** — unrelated nit, fine either way.

### Tests
Coverage of the two key code paths is good and the assertions are specific (visibility default, parentAssetId wiring, association payload shape). No edge-case tests for: `writeGpsTrack: false` with a non-empty track, `writeMetadataArtifact: false` with the adapter present, support files with explicit `visibility: 'visible'`, or an adapter throwing mid-loop after the primary upsert. Not blockers, but cheap to add given the test fixtures already exist.

### Security
Pure helper + adapter contract; no I/O, no SQL, no untrusted input handling. Nothing to flag.

PR #1183, opened 2026-05-02
Refactor QA content integration prompts

willgriffin:
## Summary
- Route smrt-facts extraction/reconcile and smrt-content review/correction through smrt-prompts.
- Add the smrt-facts AI candidate extraction API with focused tests.
- Keep content editor facts visible/auditable when facts exist for governed content.

## Tests
- pnpm --filter @happyvertical/smrt-facts exec vitest run src/__tests__/extraction.test.ts src/__tests__/reconcile.test.ts
- pnpm --filter @happyvertical/smrt-content exec vitest run src/factual-content.test.ts src/content-prompts.test.ts src/svelte/components/ContentEditor.test.ts
- git diff --check


packages/facts/src/fact-evidences.ts
```diff
@@ -0,0 +1,209 @@
...
+    const deletedEvidenceIds: string[] = [];
+    const sourceKeys = new Set(
+      sources.map((source) => `${source.sourceKind}:${source.sourceId}`),
+    );
+
+    if (sourceKeys.size === 0) {
+      return { deletedEvidenceIds };
+    }
+
+    const entries = await this.list({});
```
willgriffin: **High — cross-tenant scan / potential cross-tenant delete.**

`this.list({})` loads every evidence row in the table with no tenant predicate, then filters in application code by `sourceKind:sourceId` + `metadata.generatedBy` + `metadata.contentId`. Two problems:

1. **Scale:** this becomes O(N) over all tenants' evidence on every fact-audit repair. Easy to miss in tests because in-memory SQLite returns instantly.
2. **Cross-tenant deletion risk:** nothing in the loop checks `entry.tenantId`. If two tenants happen to have the same `(sourceKind, sourceId)` pair and both ran a fact audit on content with the same id (e.g. a shared id space, or attacker-influenced inputs), tenant A could delete tenant B's evidence.

The call site in `Content.extractReferenceFactsForAudit` (`content.ts:2008`) already has `this.tenantId` available — push it through as an option:

```ts
async replaceGeneratedForSources(
  sources: Array<{ sourceKind: string; sourceId: string }>,
  options: { generatedBy?: string; contentId?: string; tenantId?: string | null } = {},
) {
  // …
  const entries = await this.list({
    where: { tenantId: options.tenantId ?? null },
  });
```

and at the call site:

```ts
{ generatedBy: FACT_AUDIT_GENERATED_BY, contentId: this.id as string, tenantId: this.tenantId }
```

packages/content/src/content.ts
```diff
@@ -1546,6 +1774,1205 @@ ${correctedText}
...
+        metadata.factAudit.generatedBy === FACT_AUDIT_GENERATED_BY
+      ) {
+        const { factAudit: _removed, ...preservedMetadata } = metadata;
+        link.setMetadata?.(preservedMetadata);
+        await link.save();
+      }
+    }
+
+    const generatedEvidence = await evidences.list(
+      this.tenantId ? { where: { tenantId: this.tenantId } } : {},
```
willgriffin: **High — null `tenantId` silently disables the filter, fetching all tenants' rows.**

```ts
const generatedEvidence = await evidences.list(
  this.tenantId ? { where: { tenantId: this.tenantId } } : {},
);
```

When `this.tenantId` is `null` (global content, which is a valid scope per `@TenantScoped({ mode: 'optional' })`), the ternary collapses to `list({})` and you scan every evidence row across every tenant before filtering by `metadata.contentId === this.id`. Concurrent global-content audits will then race on the same set.

`null` is meaningful here — "global tenant" — and the SQL layer treats `{ where: { tenantId: null } }` as `tenant_id IS NULL` (see `FactCollection.findGlobal`). Use that:

```ts
const generatedEvidence = await evidences.list({
  where: { tenantId: this.tenantId ?? null },
});
```

Same pattern repeats in this PR at:
- `clearGeneratedFactSourcesForSources` (~line 1968)
- `getCurrentFactAuditSupportCandidates`
- `getFactAuditState`

All should use `?? null` rather than the `tenantId ? ... : {}` ternary.

packages/content/src/content-prompts.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  definePrompt,
+  type ResolvedPromptAI,
+} from '@happyvertical/smrt-prompts';
+
+export const smrtContentReviewPrompt = definePrompt({
+  key: 'smrtContent.review',
+  template: '{reviewPrompt}',
```
willgriffin: **Medium — default template ignores 6 of the 7 variables passed by callers.**

The default template here is just `'{reviewPrompt}'`, but the call site at `content.ts:3279` passes seven variables (`contentBody`, `contentDescription`, `contentId`, `contentTitle`, `kind`, `policyKey`, `reviewPrompt`). The extras are silently dropped unless an operator overrides the template — but with `editable: { template: true, params: true }` the UI invites operators to do exactly that.

Two problems:
1. The default template is a no-op wrapper around `buildContentReviewPrompt`'s output — operators editing the override see seven placeholders advertised but won't know that the rendered content already lives inside `{reviewPrompt}`.
2. Stored `metadata.prompt` (`content.ts:3331`) is just `reviewPrompt` text; the structured fields are never reachable from the audit trail.

Either include the structured fields directly in the default template (e.g. `Title: {contentTitle}\nDescription: {contentDescription}\n\n{reviewPrompt}`) so overrides have something to extend, or drop the redundant variables from the `resolvePrompt` call so operators don't see misleading placeholders.

The test at `content-prompts.test.ts:29` only asserts the override path — there's no coverage that the default template behaves sensibly.

willgriffin (commented):
Review of `refactor/qa-content-integration`. Three concerns flagged inline; the first two are cross-tenant correctness issues introduced by the new fact-audit repair paths, the third is a design smell in the new review prompt. The rest of the prompt-registry refactor and the new extraction tests look solid.

Not flagged here: the fact-audit `+server.ts` files do not check `locals` for auth, but those are auto-generated and the project enforces auth in `hooks.server.ts`, so it's a project-wide pattern rather than a regression in this PR. The bare `JSON.parse` in `parseFactExtractionResponse` / `parseClaimSupportResponse` is caught by the upstream try/catch wrapper; only the error message is degraded, so I left it off.

PR #1186, opened 2026-05-08
feat(social): add Garrula social primitives

willgriffin:
## Summary
- Adds reusable smrt-social account, post, OAuth state, analytics snapshot, server setup, and Svelte settings primitives for Garrula social publishing.
- Moves social credentials toward secret references, with explicit tenant secret store/retrieve helpers in smrt-secrets for OAuth callbacks outside ambient tenancy.
- Adds platform-scoped SocialAccount slugs so the same handle on X and YouTube cannot overwrite each other.
- Adds publish readiness helpers, publish modes, lifecycle helpers, and JSON raw analytics snapshot storage.

## Review Notes
- This intentionally lands the broader smrt-social V1 work and the slug collision fix in one release.
- Raw analytics snapshots are stored as JSON, and credential retrieval now surfaces secret/decryption problems instead of silently returning null.
- The pnpm lockfile diff is limited to the smrt-social importer changes.

## Validation
- pnpm --dir packages/social test
- pnpm --dir packages/social build
- pnpm --dir packages/social check
- pnpm --dir packages/secrets test
- pnpm --dir packages/secrets build
- node ../../scripts/verify-package-types-exports.js .
- pre-push lefthook format-check
- pre-push lefthook validate-all-workflows

### DIFF
--- a/packages/social/src/collections/oauth-state-collection.ts
+++ b/packages/social/src/collections/oauth-state-collection.ts
@@ -9,8 +9,10 @@ export class OAuthStateCollection extends SmrtCollection<OAuthState> {
   }
 
   async findExpired(now: Date = new Date()): Promise<OAuthState[]> {
-    const states = await this.list({});
-    return states.filter((state) => state.expiresAt.getTime() <= now.getTime());
+    return this.list({
+      where: { 'expiresAt <=': now },
+      orderBy: 'expiresAt ASC',
+    });
   }
 
   async deleteExpired(now: Date = new Date()): Promise<number> {
--- a/packages/social/src/collections/oauth-state-collection.ts
+++ b/packages/social/src/collections/oauth-state-collection.ts
@@ -1,0 +1,21 @@
+import { SmrtCollection } from '@happyvertical/smrt-core';
+import { OAuthState } from '../oauth-state.js';
+
+export class OAuthStateCollection extends SmrtCollection<OAuthState> {
+  static readonly _itemClass = OAuthState;
+
+  async findByState(state: string): Promise<OAuthState | null> {
+    return this.get({ state });
+  }
+
+  async findExpired(now: Date = new Date()): Promise<OAuthState[]> {
+    const states = await this.list({});
+    return states.filter((state) => state.expiresAt.getTime() <= now.getTime());
+  }
+
+  async deleteExpired(now: Date = new Date()): Promise<number> {
+    const expired = await this.findExpired(now);
+    await Promise.all(expired.map((state) => state.delete()));
+    return expired.length;
+  }
+}
--- a/packages/social/src/collections/social-post-collection.ts
+++ b/packages/social/src/collections/social-post-collection.ts
@@ -58,17 +58,24 @@ export class SocialPostCollection extends SmrtCollection<SocialPost> {
   }
 
   async findDueForPublish(now: Date = new Date()): Promise<SocialPost[]> {
-    const posts = await this.list({
-      where: {},
-      orderBy: 'scheduledAt ASC',
-    });
+    const [approved, scheduled] = await Promise.all([
+      this.list({
+        where: { status: 'approved' },
+        orderBy: 'scheduledAt ASC',
+      }),
+      this.list({
+        where: { status: 'scheduled', 'scheduledAt <=': now },
+        orderBy: 'scheduledAt ASC',
+      }),
+    ]);
 
-    return posts.filter((post) => {
-      if (post.status !== 'approved' && post.status !== 'scheduled') {
-        return false;
-      }
-      return !post.scheduledAt || post.scheduledAt.getTime() <= now.getTime();
-    });
+    return [
+      ...approved.filter((post) => post.isDueForPublish),
+      ...scheduled,
+    ].sort(
+      (a, b) =>
+        (a.scheduledAt?.getTime() ?? 0) - (b.scheduledAt?.getTime() ?? 0),
+    );
   }
 
   async recordPublishSuccess(
@@ -78,8 +85,9 @@ export class SocialPostCollection extends SmrtCollection<SocialPost> {
     const status = data.status ?? 'published';
     post.platformPostId = data.platformPostId;
     post.platformUrl = data.platformUrl;
-    post.publishedAt =
-      data.publishedAt ?? (status === 'published' ? new Date() : null);
+    if (status === 'published') {
+      post.publishedAt = data.publishedAt ?? post.publishedAt ?? new Date();
+    }
     post.status = status;
     post.errorMessage = null;
     await post.save();
@@ -91,7 +99,7 @@ export class SocialPostCollection extends SmrtCollection<SocialPost> {
     error: Error | string,
   ): Promise<SocialPost> {
     post.status = 'failed';
-    post.errorMessage = error instanceof Error ? error.message : error;
+    post.errorMessage = formatPublishError(error);
     await post.save();
     return post;
   }
@@ -109,4 +117,20 @@ export class SocialPostCollection extends SmrtCollection<SocialPost> {
     await post.save();
     return post;
   }
+}
+
+function formatPublishError(error: Error | string): string {
+  if (typeof error === 'string') {
+    return error;
+  }
+
+  const details = [error.message];
+  const code = (error as Error & { code?: unknown }).code;
+  if (code !== undefined) {
+    details.push(`code=${String(code)}`);
+  }
+  if (error.stack) {
+    details.push(error.stack);
+  }
+  return details.join('\n');
 }
--- a/packages/social/src/collections/social-post-collection.ts
+++ b/packages/social/src/collections/social-post-collection.ts
@@ -1,0 +1,112 @@
+import { SmrtCollection } from '@happyvertical/smrt-core';
+import {
+  type PostAnalytics,
+  type PostStatus,
+  SocialPost,
+  type SocialPostType,
+} from '../social-post.js';
+
+export interface CreateSocialPostDraftOptions {
+  socialAccountId: string;
+  contentId?: string | null;
+  videoContentId?: string | null;
+  postType?: SocialPostType;
+  title?: string | null;
+  description?: string;
+  hashtags?: string[];
+  linkUrl?: string | null;
+  mediaUrl?: string | null;
+  scheduledAt?: Date | null;
+  tenantId?: string | null;
+}
+
+export interface PublishSuccessData {
+  platformPostId: string;
+  platformUrl: string;
+  publishedAt?: Date;
+  status?: PostStatus;
+}
+
+export class SocialPostCollection extends SmrtCollection<SocialPost> {
+  static readonly _itemClass = SocialPost;
+
+  async createDraft(
+    options: CreateSocialPostDraftOptions,
+  ): Promise<SocialPost> {
+    return this.create({
+      ...options,
+      postType:
+        options.postType ??
+        (options.videoContentId || options.mediaUrl
+          ? 'video'
+          : options.linkUrl
+            ? 'link'
+            : 'text'),
+      status: options.scheduledAt ? 'scheduled' : 'draft',
+    });
+  }
+
+  async findByStatus(status: PostStatus): Promise<SocialPost[]> {
+    return this.list({ where: { status }, orderBy: 'scheduledAt ASC' });
+  }
+
+  async findForAccount(socialAccountId: string): Promise<SocialPost[]> {
+    return this.list({
+      where: { socialAccountId },
+      orderBy: 'created_at DESC',
+    });
+  }
+
+  async findDueForPublish(now: Date = new Date()): Promise<SocialPost[]> {
+    const posts = await this.list({
+      where: {},
+      orderBy: 'scheduledAt ASC',
+    });
+
+    return posts.filter((post) => {
+      if (post.status !== 'approved' && post.status !== 'scheduled') {
+        return false;
+      }
+      return !post.scheduledAt || post.scheduledAt.getTime() <= now.getTime();
+    });
+  }
+
+  async recordPublishSuccess(
+    post: SocialPost,
+    data: PublishSuccessData,
+  ): Promise<SocialPost> {
+    const status = data.status ?? 'published';
+    post.platformPostId = data.platformPostId;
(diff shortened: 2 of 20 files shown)


packages/social/src/social-account.ts
```diff
@@ -279,8 +416,122 @@ export class SocialAccount extends SmrtObject {
...
+      : await this.withCredentialTenantContext(() =>
+          secretService.retrieve(credentialSecretId),
+        );
+    return JSON.parse(secret.value);
+  }
+
+  private async withCredentialTenantContext<T>(
+    fn: () => Promise<T>,
+  ): Promise<T> {
+    if (getCurrentTenant()) {
```
willgriffin: **Cross-tenant divergence.** This checks for *any* ambient tenant, not whether ambient matches `this.tenantId`. If ambient is tenant A but `this.tenantId === 'B'`, the secret was just written under B (via `storeForTenant` above) but the followup `save()` runs under A. That's a silent inconsistency.

Either assert ambient matches `this.tenantId` and throw on mismatch, or always run the save under `this.tenantId` when set. The same shape applies to `getCredentials`: if ambient and account tenants drift, you can read the wrong secret because disambiguation depends on whichever tenant ALS happened to have, not on the model.

packages/social/src/social-account.ts
```diff
@@ -242,17 +331,60 @@ export class SocialAccount extends SmrtObject {
...
+    if (options.publicPublishingAllowed !== undefined)
+      this.publicPublishingAllowed = options.publicPublishingAllowed;
     if (options.status !== undefined) this.status = options.status;
     if (options.errorMessage !== undefined)
       this.errorMessage = options.errorMessage;
-    if (options.tenantId !== undefined) this.tenantId = options.tenantId;
+    if (options.tenantId !== undefined) {
+      this.tenantId = options.tenantId;
+    } else {
+      const rawTenantId = (options as { tenant_id?: string | null }).tenant_id;
```
willgriffin: **Snake-case `tenant_id` fallback only on `SocialAccount`.** No other tenant-scoped model in the repo (Membership, Group, Role, Session, OAuthState, SocialPost, SocialPostAnalyticsSnapshot, …) has this pattern, and `SmrtObject.loadDataFromDb()` already rekeys snake → camel before hydrating. Two reads:

- It's defending against an upstream caller that hands a raw row through the constructor — fix the upstream, don't patch one model.
- Or it's dead code from an earlier prototype.

Either way, pick one and leave a comment explaining the call site. The cast-and-grab pattern silently erodes typing.

Same concern at line 532 in `getCredentialTenantId`.

willgriffin (commented):
Deep review — net positive PR. Modeling, types, and the publish-mode safety story are good.

Recommend addressing the cross-tenant divergence, the `tenant_id` snake_case fallback, the double-save in `setCredentials`, and the optional-tenancy `setCredentials` path before merge. The N+1-style `findDueForPublish`/`findExpired` and collection test coverage can be a fast-follow if there's pressure to land V1.

### Higher-level items not pinned to a single line

**Coverage gaps on the new collections.** No tests for `OAuthStateCollection.{findExpired,deleteExpired}`, `SocialAccountCollection.{findActive,findReady,findNeedsAttention}`, `SocialPostAnalyticsSnapshotCollection.{recordSnapshot,findForPost,findLatestForPost}`, or `SocialPostCollection.{createDraft,findDueForPublish,recordPublishFailure,updateLatestAnalytics}`. With the in-memory DB helper (`createIsolatedTestDbFromManifest({ includeObjects: [...] })`) these are easy to write and would also flush out the cross-tenant + optional-tenancy edge cases below.

**`SOCIAL_PLATFORM_SETUPS.defaultPublishMode` / `defaultLinkBehavior` are never applied.** They're data the consumer has to wire up. Either export an `applyPlatformDefaults(account)` helper that uses them, or drop the per-platform default and rely on the model defaults — otherwise a caller can easily forget to read them and ship `dry_run` for every platform.

**End-to-end coverage for `setCredentials` / `getCredentials`.** A round-trip test with a real `SecretService` + in-memory DB would lock down the cross-tenant behavior flagged below.

**Misc nits not worth pinning:** `recordPublishFailure` only stores `error.message` — no stack/code; consider stuffing the stack into a metadata field for triage. `getSocialPlatformSetup` is O(n) over 5 entries (fine), convert to a `Map` if the platform list ever grows.

packages/social/src/social-account.ts
```diff
@@ -279,8 +416,122 @@ export class SocialAccount extends SmrtObject {
...
+   * Store all platform credentials in smrt-secrets as a single JSON payload.
+   */
+  async setCredentials(
+    credentials: Record<string, unknown>,
+    options: {
+      description?: string;
+      category?: string;
+    } = {},
+  ): Promise<void> {
+    if (!this.id) {
```
willgriffin, line 469: **Two writes per first-time `setCredentials`.** This branch saves once before knowing `credentialSecretId`, then saves again at line 483 inside `withCredentialTenantContext` to persist the secret name. Two upserts, two interceptor passes, two audit rows (depending on hooks).

You can compute `secretName` from `this.id ?? crypto.randomUUID()` (or `await this.getId()` which already mints one), set `this.id` and `this.credentialSecretId` *before* the first save, and write once.

packages/social/src/social-account.ts
```diff
@@ -279,8 +416,122 @@ export class SocialAccount extends SmrtObject {
...
+        tenantId,
+        secretName,
+        JSON.stringify(credentials),
+        {
+          description: options.description ?? `Credentials for ${this.name}`,
+          category: options.category ?? 'social',
+        },
+      );
+    } else {
+      await secretService.store(secretName, JSON.stringify(credentials), {
```
willgriffin: **`setCredentials` with optional-tenancy + no ambient context fails opaquely.** This else-branch calls `secretService.store(...)` which calls `requireTenantId()` and throws when there's no ambient ALS tenant. A `SocialAccount` declared `TenantScoped({ mode: 'optional' })` with `tenantId = null` and no surrounding `withTenant()` will hit this with a confusing error.

Either allow `storeForTenant(null, ...)` in optional mode (i.e. accept "global"), or throw a clear "tenant context required to store credentials" error before falling into `SecretService`.

packages/social/src/social-account.ts
```diff
@@ -279,8 +416,122 @@ export class SocialAccount extends SmrtObject {
...
+    options: {
+      description?: string;
+      category?: string;
+    } = {},
+  ): Promise<void> {
+    if (!this.id) {
+      await this.save();
+    }
+
+    const { SecretService } = await import('@happyvertical/smrt-secrets');
```
willgriffin: **Dynamic import is unnecessary.** `@happyvertical/smrt-secrets` is now in `dependencies` (this PR added it). Static `import { SecretService } from '@happyvertical/smrt-secrets'` works and gives better tree-shaking + types. The dynamic form looks like vestigial defensive code from when this was speculative. Same at line 499.

packages/social/src/social-account.ts
```diff
@@ -242,17 +331,60 @@ export class SocialAccount extends SmrtObject {
...
+   * Social accounts need a slug identity that is scoped by tenant and platform.
+   * A newsroom may connect `@localnews` on X, YouTube, Threads, and Facebook;
+   * the generic name-derived slug would make those accounts overwrite each
+   * other through SMRT's slug/context upsert identity.
+   */
+  async getSlug(): Promise<string | null | undefined> {
+    if (!this.slug) {
+      const identity =
+        this.platformUserId || this.platformUsername || this.name || this.id;
+      const source = [this.tenantId || 'global', this.platform, identity]
```
willgriffin: **`'global'` sentinel can collide with a real tenant.** If a tenant is ever literally named `global`, its accounts share a slug pool with truly-untenanted accounts. Consider `'tenant-null'` or `_global_`, or otherwise a sentinel that can't be a tenant id. Edge case.

packages/social/src/collections/social-post-collection.ts
```diff
@@ -0,0 +1,112 @@
...
+  async findForAccount(socialAccountId: string): Promise<SocialPost[]> {
+    return this.list({
+      where: { socialAccountId },
+      orderBy: 'created_at DESC',
+    });
+  }
+
+  async findDueForPublish(now: Date = new Date()): Promise<SocialPost[]> {
+    const posts = await this.list({
+      where: {},
```
willgriffin: **Loads the entire posts table on every scheduler tick.** `where: {}` returns every row, then JS filters by status + scheduledAt. The collection's `where` supports operator suffixes (`'status in'`, `'scheduledAt <='`), so push these down. The OR-with-null on `scheduledAt` is awkward but can be split into two `list()` calls and merged, or use `query()` with raw SQL. As-is this scales linearly with the table.

packages/social/src/collections/social-post-collection.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  async recordPublishSuccess(
+    post: SocialPost,
+    data: PublishSuccessData,
+  ): Promise<SocialPost> {
+    const status = data.status ?? 'published';
+    post.platformPostId = data.platformPostId;
+    post.platformUrl = data.platformUrl;
+    post.publishedAt =
```
willgriffin: **`publishedAt` is unconditionally rewritten.** If a post was previously `published` (so it has a `publishedAt`) and is then re-recorded with `status: 'staged'` or `'dry_run'`, the original timestamp is wiped. Probably an unusual flow, but consider preserving the previous timestamp when the new status isn't `published`.

packages/social/src/collections/oauth-state-collection.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class OAuthStateCollection extends SmrtCollection<OAuthState> {
+  static readonly _itemClass = OAuthState;
+
+  async findByState(state: string): Promise<OAuthState | null> {
+    return this.get({ state });
+  }
+
+  async findExpired(now: Date = new Date()): Promise<OAuthState[]> {
+    const states = await this.list({});
```
willgriffin: **Loads every OAuth state row to find expired ones.** `await this.list({})` materializes the whole table, then `findExpired` filters in JS. Push the `expiresAt <=` predicate down via `where: { 'expiresAt <=': now }` so the DB does the work. Same applies to `deleteExpired`, which then iterates and calls `delete()` per-row — fine for a few rows but could be a single SQL delete via `query()` if the table grows.

packages/social/src/social.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  it('does not stamp dry-run or staged posts as published', async () => {
+    const post = new SocialPost({
+      socialAccountId: 'account-1',
+      postType: 'link',
+      status: 'publishing',
+    });
+    post.save = vi.fn(async () => post);
+
+    await SocialPostCollection.prototype.recordPublishSuccess.call(
```
willgriffin: **Test bypasses the collection.** This works only because `recordPublishSuccess` happens not to use `this`. Any future maintainer who adds `this.create()` or an interceptor call will break the test in a confusing way.

Use a real (in-memory) `SocialPostCollection` per `.claude/rules/testing.md`:

```ts
const { db } = await createIsolatedTestDbFromManifest({
  includeObjects: ['SocialPost'],
});
const collection = await SocialPostCollection.create({ db });
```

packages/social/src/index.ts
```diff
@@ -54,4 +56,9 @@ export {
   type PostStatus,
   SocialPost,
   type SocialPostOptions,
+  type SocialPostType,
 } from './social-post.js';
```
willgriffin, line 60: **`RawAnalyticsPayload` is not re-exported.** Consumers calling `SocialPostAnalyticsSnapshotCollection.recordSnapshot({ raw: ... })` cannot type the argument without reaching into `./social-post-analytics-snapshot.js` directly. Add it to the barrel here.

packages/social/src/social-post-analytics-snapshot.ts
```diff
@@ -0,0 +1,85 @@
...
+  cli: true,
+})
+export class SocialPostAnalyticsSnapshot extends SmrtObject {
+  @tenantId({ nullable: true })
+  tenantId: string | null = null;
+
+  @foreignKey(() => SocialPost)
+  socialPostId: string | null = null;
+
+  platform: SocialPlatformType = 'youtube';
```
willgriffin: **Defaulting `platform` to `'youtube'` hides bugs.** A defaulted-to-arbitrary enum will silently swallow "forgot to set platform" mistakes — the snapshot will be persisted as YouTube data with whatever metrics actually came in. If the field is required in practice, drop the default and surface a validation error (or use a sentinel like `'unknown'` if the column must be non-null).

packages/social/src/server.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export function getDefaultSocialAccountName(
+  platform: SocialPlatformType,
+  username?: string | null,
+): string {
+  const setup = getSocialPlatformSetup(platform);
+  return username ? `${setup.label} ${username}` : setup.label;
+}
+
+export function normalizeManualCredentials(
```
willgriffin, line 105: **Silent on missing required fields.** This trims and includes only string values for `requiredSecretFields`. If a required field is missing or empty, it's silently omitted — no error. Caller is responsible for verifying completeness, but there's no helper to do so. Consider a companion `validateManualCredentials(platform, values)` that returns the missing list, so upstream forms can surface a useful error.

packages/social/src/svelte/components/SocialAccountSettings.svelte
```diff
@@ -0,0 +1,239 @@
...
+  { platform: 'facebook', label: 'Facebook Page' },
+  { platform: 'threads', label: 'Threads' },
+  { platform: 'x', label: 'X' },
+  { platform: 'bluesky', label: 'Bluesky' },
+];
+
+function statusLabel(account: SocialAccountSettingsItem): string {
+  if (!account.isActive) return 'inactive';
+  if (account.needsAttention) return 'needs attention';
+  return account.status.replace('_', ' ');
```
willgriffin: **`replace` only swaps the first underscore.** Line 79 in this file uses `replaceAll('_', ' ')` for `publishMode`; this one uses `replace('_', ' ')`. Use `replaceAll` for consistency. Today's only impact is on multi-underscore future statuses.

Issue #1189, opened 2026-05-08
[smrt-content] Make content editor assistant context reusable outside the local editor sidebar


willgriffin:
## Summary

Anytown is using `@happyvertical/smrt-content`'s `GovernedContentEditor`, but the current upstream chat/editor integration is still shaped around an editor-local `ContentAgentChat` sidebar. In the dashboard we now want a persistent workspace Assistant that stays mounted at the portal layout level, follows the current page, and can act on the active content editor.

This feels like it belongs upstream in `smrt-content`: the package already owns the content editor, governed editor, content chat prompts, and content chat route contracts, so consumers should not need to hand-roll page context stores and copied chat endpoints just to move the chat surface.

## Downstream evidence in anytown.ai

Current local pieces that look upstream-worthy:

- `apps/dashboard/src/routes/sites/[siteSlug]/articles/[articleId]/+page.svelte`
  - Renders `GovernedContentEditor` with `hideChat={true}`.
  - Registers page/editor context into `@anytown/site-portal`'s global `assistantStore`.
  - Exposes editor actions through component refs: `triggerSave()` and `triggerReview(kind)`.
- `packages/site-portal/src/lib/stores/assistant.svelte.ts`
  - Defines a generic assistant context shape with `type`, `title`, serializable `data`, and local `actions`.
- `packages/site-portal/src/lib/components/PortalAssistantSidebar.svelte`
  - Mounts a persistent Assistant at the layout level and sends serialized context with each chat message.
- `apps/dashboard/src/routes/sites/[siteSlug]/api/v1/contents/[id]/chat/**`
  - Locally mirrors/adapts the `smrt-content` content chat endpoints so Anytown can add site tenancy, authenticated profile checks, base-path routing, and Bifrost-compatible AI/model resolution.

There is also an older local `apps/dashboard/src/lib/components/editor/RichEditor.svelte` / `WritingAssistant.svelte` pair with quick writing actions, dictation, and read-aloud controls. That appears less central than the active `GovernedContentEditor` path, but it points at a future extension point for optional editor toolbar actions.

## Proposed upstream feature

Add first-class support for a content-editor assistant context that can be consumed by either the built-in sidebar or an application-level assistant shell.

Possible API shape:

- A serializable `ContentEditorAssistantContext` type containing:
  - `contentId`
  - content/editor type and title
  - current draft fields used by content chat (`title`, `description`, `type`, `status`, `state`, `publish_date`, `body`)
  - `currentEditorState`
[...]

PR #1190, opened 2026-05-09
Add content editor WYSIWYG body and assistant context

willgriffin:
## Summary
- add `bodyFormat` plumbing, body-format utilities, a native WYSIWYG body editor, and Markdown/HTML body rendering
- support inline body images from chooser/drag payloads, primary-image selection from the first inserted image, and carousel selection for body images
- fold in the content assistant-context work for reusable editor context/action types, external chat context, shared server chat handlers, and route refactors

## Review cleanup
- fixed assistant context registration so editor field changes do not transiently unregister context
- made body/Markdown renderer props reactive for Svelte updates
- added package build entries for body-format and assistant-context modules so packed Svelte exports resolve
- removed the remaining non-null assertion warning from the chat prompt test

## Verification
- `pnpm --filter @happyvertical/smrt-content test -- body-format.test.ts ContentEditor.test.ts ContentAgentChat.test.ts content-chat-prompts.test.ts`
- `pnpm --filter @happyvertical/smrt-content check`
- `pnpm --filter @happyvertical/smrt-content build`
- `pnpm --filter @happyvertical/smrt-content verify:pack`
- `git diff --check`

Refs #1189


packages/content/src/body-format.ts
```diff
@@ -0,0 +1,676 @@
...
+      doubleValue = '',
+      singleValue = '',
+      bareValue = '',
+    ) => {
+      const safeStyle = sanitizeStyle(doubleValue || singleValue || bareValue);
+      return safeStyle ? ` style="${escapeAttribute(safeStyle)}"` : '';
+    },
+  );
+  html = html.replace(
+    /\s+(href|src)\s*=\s*("([^"]*)"|'([^']*)'|([^\s>]+))/gi,
```
willgriffin: **P0 — sanitizer bypass: `xlink:href` (and `formaction`/`action`/`poster`) not stripped**

This regex only matches `href` and `src`. SVG namespaced attributes pass through unchanged, so a stored body containing

```html
<svg><a xlink:href="javascript:alert(1)">click</a></svg>
```

renders verbatim via `{@html safeHtml}` in [`ContentBodyRenderer.svelte:25`](../blob/74e2223024036ebab09ecc842f1ae03466897bbe/packages/content/src/svelte/components/ContentBodyRenderer.svelte#L25) and executes on click. `formaction` on `<button>` and `poster` on `<video>` are similar vectors.

**Fix:** broaden the attribute list and run all URL-bearing attributes through `sanitizeUrl`:

```ts
html = html.replace(
  /\s+(href|src|xlink:href|formaction|action|poster)\s*=\s*("([^"]*)"|'([^']*)'|([^\s>]+))/gi,
  ...
);
```

Also please add regression tests in `body-format.test.ts` covering each of these attribute names. Long-term, the regex pipeline is fragile — moving to a parser-based allowlist sanitizer (e.g. DOMPurify) would close this whole class of bypass.

packages/content/src/content-chat-handlers.ts
```diff
@@ -0,0 +1,569 @@
...
+    const id = asNonEmptyString(refId);
+    if (!id) {
+      continue;
+    }
+    const ref = (await contents.get(id)) as {
+      title?: string;
+      body?: string;
+    } | null;
+    if (ref) {
+      refTexts.push(`--- REFERENCE: ${ref.title} ---\n${ref.body}`);
```
willgriffin: **P0 — prompt injection via reference body**

`ref.title` and `ref.body` are concatenated verbatim into the AI system prompt with no length cap, no HTML stripping, and no escaping. Any user who can author a content record (which is the entire intended use case) can plant prompt-injection text — `IGNORE PREVIOUS INSTRUCTIONS...` style — that gets pulled into other authors' chat sessions whenever their content references it.

`stripHtml` is already exported from `body-format.ts`; use it plus a per-reference cap and a clearer delimiter:

```ts
const safeTitle = stripHtml(ref.title ?? '').slice(0, 200);
const safeBody = stripHtml(ref.body ?? '').slice(0, 2000);
refTexts.push(
  `<<<REFERENCE id=${id}>>>\nTitle: ${safeTitle}\nBody: ${safeBody}\n<<<END REFERENCE>>>`,
);
```

Using distinctive delimiters (rather than `--- REFERENCE ---`, which an attacker can replicate) makes it harder for injected content to claim authority over the surrounding prompt.

packages/content/src/svelte/components/GovernedContentEditor.svelte
```diff
@@ -141,6 +149,58 @@ const publishSaveDisabled = $derived(
...
+});
+
+$effect(() => {
+  const callback = onAssistantContextChange;
+  if (!callback) {
+    return;
+  }
+
+  return () => callback(null);
+});
```
willgriffin, line 249: **P1 — split `$effect`s can transiently unregister context**

The PR description says "fixed assistant context registration so editor field changes do not transiently unregister context." The remaining structure still has the same hazard for any consumer that passes an inline arrow as `onAssistantContextChange`:

- Effect 1 (lines 192–194) re-runs on registration changes and pushes the new value.
- Effect 2 (lines 196–203) only depends on `onAssistantContextChange`. When the parent re-renders with a new inline callback reference, Effect 2 tears down — firing `callback(null)` against the previous callback — and re-runs.

The order between Effect 1 and Effect 2's cleanup isn't guaranteed, so consumers can observe `register → null` or `null → register` either way. The same dual-effect pattern is in [`ContentEditor.svelte:209-223`](../blob/74e2223024036ebab09ecc842f1ae03466897bbe/packages/content/src/svelte/components/ContentEditor.svelte#L209-L223) and has the same issue.

**Fix:** collapse into one effect so the register and unregister are paired atomically:

```ts
$effect(() => {
  const callback = onAssistantContextChange;
  if (!callback) return;
  callback(governedAssistantRegistration);
  return () => callback(null);
});
```

This fires the new registration first, and the cleanup only runs on actual unmount or when the dependencies change — and even then, in cleanup-then-body order on the same callback reference.

packages/content/src/svelte/components/ContentEditor.svelte
```diff
@@ -35,6 +50,7 @@ let {
   agentChatNotice?: string | null;
   hideActions?: boolean;
   hideChat?: boolean;
+  onAssistantContextChange?: ContentEditorAssistantContextChange;
```
willgriffin: **P1 — `$props<{...}>()` inline-generic form violates project convention**

CLAUDE.md flags this exact pattern as a TypeScript/Svelte 5 perf hazard:

> Avoid using inline intersected generic types (like `Asset & { id: string }`) in component `$props()`. It causes infinite-loop-like recursion during type evaluation. Export an explicit interface instead.

Every other new component in this PR (`GovernedContentEditor`, `ContentBodyEditor`, `ContentBodyRenderer`, `Markdown`) correctly uses `export interface Props { ... }` + `let { ... }: Props = $props()`. `ContentEditor` is the odd one out — please align:

```ts
export interface Props {
  apiBaseUrl?: string;
  content?: any;
  // ...
  onSave: (data: any) => void;
  onCancel: () => void;
}

let { apiBaseUrl = '/api/v1', /* ... */ }: Props = $props();
```

(Same applies to `ContentAgentChat.svelte:32-41`, though that one predates this PR — but you're materially expanding it here, so worth folding in.)

willgriffin (commented):
Reviewed the WYSIWYG body editor, body-format sanitizer, assistant-context plumbing, and chat handler refactor.

Four findings worth addressing — two security (P0), two correctness/convention (P1). Posting as inline comments below. Overall the structure is sound; the sanitizer surface and the dual-`$effect` registration pattern are the main concerns.

Not included as separate comments but worth tracking:
- The sanitizer is regex-based across HTML, attributes, styles, and URLs. Long-term, swapping to a parser-based allowlist sanitizer (DOMPurify or similar) would be safer than playing whack-a-mole with regex bypasses. Each new vector (xlink:href today, something else next quarter) requires a new regex.
- `body-format.test.ts` covers `javascript:` URLs and `on*=` handlers but has no test for `xlink:href`, `formaction`, or `<svg>`-namespaced vectors. After fixing #1, please add regression tests for these specific bypass patterns.
- `document.execCommand` (ContentBodyEditor.svelte:562) is deprecated; not a bug today, but Chrome has signaled intent to remove. Worth a tracking issue.
- `currentProfileId = 'user123'` hardcode in `ContentAgentChat.svelte:52` is pre-existing (not introduced here), but since this PR materially extends ContentAgentChat into a standalone-mountable component, it'd be a good time to make it a real prop driven by auth.

PR #1201, opened 2026-05-09
feat(languages): add @happyvertical/smrt-languages package

willgriffin:
## Summary

Closes #1200. Adds `@happyvertical/smrt-languages` — a new foundational package that mirrors the architecture of `@happyvertical/smrt-prompts` for user-facing strings, plus an AI auto-translation pipeline backed by `smrt-jobs` that fills missing locales the first time a string is requested.

- Code-first defaults via `defineLanguageString({ key, locale, template })`
- 5-layer resolution: code → file config → app override → tenant override → runtime override
- BCP-47 locale fallback chain (`fr-CA` → `fr` → registered default)
- Soft-miss returns the fallback immediately and enqueues an AI translation job; deterministic dedup ID `smrt-languages.translate:<key>:<locale>` so concurrent misses collapse into one job
- `LanguageOverride` rows on `_smrt_language_overrides` carry `auto_generated`, `source_hash`, `ai_model`, `reviewed_at`, `reviewed_by` for the admin review queue
- Honors a `smrt-features` kill switch (`smrt-languages.auto_translate`), an optional `supportedLocales` allowlist, and a per-tenant daily budget
- Source-hash gating: never re-translates when the source hasn't changed; never overwrites a row with `auto_generated: false` (human edits win permanently)
- Translation prompt itself registered with `smrt-prompts` under `smrt-languages.translation` so ops can tune wording without redeploying
- Tenant glossary (existing tenant overrides) injected into the AI prompt so auto-translations match tenant voice

This is a Phase 2 prerequisite for the per-package adoption epic (#1199 / #1191).

## Test plan

- [x] `pnpm --filter @happyvertical/smrt-languages test` — 30 tests across 4 files: utils (15), resolver (9), translation-job dedup (4), integration (2)
- [x] Integration test: register `en` default → resolve in `es` (miss → fallback to `en`, returns `source: 'fallback'`) → run the task's `execute` directly with mocked `@happyvertical/ai` → verify app override row persists with `auto_generated: true` + `source_hash` → resolve again in `es` returns the translated string with `source: 'app'`
- [x] Integration test: human-edited row (`auto_generated: false`) is preserved when the source changes — handler returns `skipped: 'stale'` and the original template stays put
- [x] `pnpm --filter @happyvertical/smrt-languages build` — vite library build + DTS rollup, manifest with 3 objects
- [x] `npx biome check packages/languages/src` — clean
- [x] Independent code review (review pass found 2 real issues — both fixed: `translateMissing` now skips human-edited rows before enqueue (`packages/languages/src/cli.ts`), and the dedup scan logs a warning when truncated and accepts a tunable `dedupScanLimit` config)

## Notes

- The `LanguageTranslationTask` SmrtObject is registered via `@smrt()` so the existing `TaskRunner` resolves jobs by `objectType` and invokes `execute(args, executionContext)` — the standard contract.
- The package is added to the `.changeset/config.json` fixed group so its version stays locked with the rest of the workspace.
- CLI helpers (`translateMissing`, `listUnreviewedAutoTranslations`, `approveAutoTranslation`, `editLanguageOverride`) ship as exported functions in [src/cli.ts](packages/languages/src/cli.ts) — wiring them into the actual `smrt languages …` subcommand surface in `packages/cli` is a small follow-up that touches the CLI generator, not this package.

### DIFF
--- a/packages/languages/src/index.ts
+++ b/packages/languages/src/index.ts
@@ -1,0 +1,73 @@
+/**
+ * @happyvertical/smrt-languages
+ *
+ * Code-first language strings with config + tenant overrides and AI-driven
+ * auto-translation for SMRT applications. Mirrors the architecture of
+ * `@happyvertical/smrt-prompts` plus a smrt-jobs handler that fills gaps in
+ * non-default locales the first time a string is requested.
+ *
+ * @packageDocumentation
+ */
+
+import './__smrt-register__.js';
+
+export {
+  clearLanguageCache,
+  getLanguageCacheTtlMs,
+  invalidateLanguageCache,
+} from './cache.js';
+export {
+  approveAutoTranslation,
+  editLanguageOverride,
+  listUnreviewedAutoTranslations,
+  translateMissing,
+} from './cli.js';
+
+export { LanguageOverrideCollection } from './collections/LanguageOverrideCollection.js';
+
+export { buildTenantGlossary } from './glossary.js';
+
+export {
+  defineLanguageString,
+  LanguageRegistry,
+} from './language-registry.js';
+
+export {
+  invalidateResolvedLanguageCache,
+  resolveLanguageString,
+} from './language-resolver.js';
+
+export {
+  LanguageOverride,
+  type LanguageOverrideCtorOptions,
+} from './models/LanguageOverride.js';
+
+export {
+  AUTO_TRANSLATE_FEATURE_KEY,
+  enqueueTranslationJob,
+  LanguageTranslationTask,
+  TRANSLATION_PROMPT_KEY,
+} from './translation-job.js';
+
+export type {
+  LanguageCacheValue,
+  LanguageOverrideOptions,
+  LanguageStringDefinition,
+  LanguageStringDefinitionInput,
+  LanguagesPackageConfig,
+  LanguageVariables,
+  ResolvedLanguageString,
+  ResolveLanguageStringOptions,
+  TranslationJobPayload,
+} from './types.js';
+
+export {
+  buildLocaleFallbackChain,
+  buildTranslationJobId,
+  computeSourceHash,
+  normalizeLocale,
+  renderTemplate,
+} from './utils.js';
+
+/** @internal */
+export const PACKAGE_VERSION_INITIALIZED = true;
--- a/packages/languages/src/index.ts
+++ b/packages/languages/src/index.ts
@@ -6,6 +6,15 @@
  * `@happyvertical/smrt-prompts` plus a smrt-jobs handler that fills gaps in
  * non-default locales the first time a string is requested.
  *
+ * The package root intentionally exposes only the read path
+ * (`defineLanguageString`, `resolveLanguageString`, `LanguageOverride`, the
+ * cache helpers, glossary helpers, and shared types/utilities). The
+ * translation-worker stack — which depends on `@happyvertical/ai`,
+ * smrt-jobs, smrt-features, and smrt-prompts — lives on the
+ * `@happyvertical/smrt-languages/jobs` subpath, and the admin CLI helpers
+ * live on `@happyvertical/smrt-languages/cli`. Resolver consumers (the vast
+ * majority of call sites) never load the worker dependency tree.
+ *
  * @packageDocumentation
  */
 
@@ -16,15 +25,7 @@ export {
   getLanguageCacheTtlMs,
   invalidateLanguageCache,
 } from './cache.js';
-export {
-  approveAutoTranslation,
-  editLanguageOverride,
-  listUnreviewedAutoTranslations,
-  translateMissing,
-} from './cli.js';
-
 export { LanguageOverrideCollection } from './collections/LanguageOverrideCollection.js';
-
 export { buildTenantGlossary } from './glossary.js';
 
 export {
@@ -42,13 +43,6 @@ export {
   type LanguageOverrideCtorOptions,
 } from './models/LanguageOverride.js';
 
-export {
-  AUTO_TRANSLATE_FEATURE_KEY,
-  enqueueTranslationJob,
-  LanguageTranslationTask,
-  TRANSLATION_PROMPT_KEY,
-} from './translation-job.js';
-
 export type {
   LanguageCacheValue,
   LanguageOverrideOptions,
@@ -58,7 +52,6 @@ export type {
   LanguageVariables,
   ResolvedLanguageString,
   ResolveLanguageStringOptions,
-  TranslationJobPayload,
 } from './types.js';
 
 export {
--- a/packages/languages/src/language-resolver.ts
+++ b/packages/languages/src/language-resolver.ts
@@ -100,8 +100,11 @@ async function resolveBaseForLocale(
     }
   }
 
-  // 3. File-config override.
-  const configTemplate = config.overrides?.[key]?.[locale];
+  // 3. File-config override. Locale keys in user-authored config files
+  // commonly come in mixed case (`fr-ca`, `Fr-CA`, etc.); normalize lookup so
+  // config matches the same way the DB and code-default layers do.
+  const configForKey = config.overrides?.[key];
+  const configTemplate = lookupNormalizedLocale(configForKey, locale);
   if (typeof configTemplate === 'string') {
     return finalize({
       key,
@@ -137,6 +140,21 @@ async function resolveBaseForLocale(
     source: 'missing',
     resolvedFromLocale: locale,
   };
+}
+
+function lookupNormalizedLocale(
+  bag: Record<string, string> | undefined,
+  locale: string,
+): string | undefined {
+  if (!bag) return undefined;
+  // Fast path: exact match (most file configs use the canonical form).
+  if (typeof bag[locale] === 'string') return bag[locale];
+  for (const [candidate, template] of Object.entries(bag)) {
+    if (normalizeLocale(candidate) === locale) {
+      return template;
+    }
+  }
+  return undefined;
 }
 
 function finalize(args: {
@@ -230,7 +248,6 @@ export async function resolveLanguageString(
       requestedLocale: requested,
       defaultLocale,
       options,
-      fallbackTemplate: firstHit.template,
     });
     return {
       key,
@@ -265,7 +282,6 @@ async function tryEnqueueTranslation(args: {
   requestedLocale: string;
   defaultLocale: string;
   options: ResolveLanguageStringOptions;
-  fallbackTemplate?: string | null;
 }): Promise<void> {
   if (!args.options.db) {
     // Without a DB we cannot persist a job; skip silently — resolver still
--- a/packages/languages/src/language-resolver.ts
+++ b/packages/languages/src/language-resolver.ts
@@ -1,0 +1,300 @@
+import { getPackageConfig } from '@happyvertical/smrt-config';
+import { getTenantId } from '@happyvertical/smrt-tenancy';
+import {
+  getCachedLanguage,
+  invalidateLanguageCache,
+  setCachedLanguage,
+} from './cache.js';
+import { LanguageOverrideCollection } from './collections/LanguageOverrideCollection.js';
+import { LanguageRegistry } from './language-registry.js';
+import type {
+  LanguageCacheValue,
+  LanguagesPackageConfig,
+  ResolvedLanguageString,
+  ResolveLanguageStringOptions,
+} from './types.js';
+import {
+  buildLocaleFallbackChain,
+  normalizeLocale,
+  renderTemplate,
+} from './utils.js';
+
+const DEFAULT_LOCALE = 'en';
+
+function getLanguagesConfig(): LanguagesPackageConfig {
+  return getPackageConfig<LanguagesPackageConfig>('languages', {
+    defaultLocale: DEFAULT_LOCALE,
+    overrides: {},
+  });
+}
+
+interface ResolutionAttempt {
+  template: string | null;
+  source: ResolvedLanguageString['source'];
+  resolvedFromLocale: string;
+}
+
+async function resolveBaseForLocale(
+  key: string,
+  locale: string,
+  options: ResolveLanguageStringOptions,
+  config: LanguagesPackageConfig,
+): Promise<ResolutionAttempt> {
+  const tenantId =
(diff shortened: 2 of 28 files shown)


packages/languages/src/translation-job.ts
```diff
@@ -0,0 +1,401 @@
...
+export const TRANSLATION_PROMPT_KEY = 'smrt-languages.translation';
+
+definePrompt({
+  key: TRANSLATION_PROMPT_KEY,
+  template:
+    'You translate user-facing application strings from {sourceLocale} to {targetLocale}.\n' +
+    'Preserve any "{var}" placeholders exactly. Do not add markup or commentary.\n' +
+    'Match the organization glossary where applicable:\n{glossary}\n' +
+    'String to translate: "{template}"\n' +
+    'Reply with a JSON object: {"translation": string}.',
```
willgriffin: The default prompt contains literal braces, but `resolvePrompt()` renders every `{...}` as a variable. With the current variables, `"{var}"` becomes `""` and `{"translation": string}` becomes empty, so the model never sees the placeholder-preservation or JSON-shape instructions. Please avoid raw braces in prompt templates or add an escaping/literal mechanism before this ships, otherwise translations can drop `{name}` placeholders or return a shape the parser was not actually told to use.

packages/languages/src/translation-job.ts
```diff
@@ -0,0 +1,401 @@
...
+  try {
+    const jobs = await (SmrtJobCollection as any).create({ db: options.db });
+    const since = new Date();
+    since.setUTCHours(0, 0, 0, 0);
+    const rows = await jobs.list({
+      where: {
+        objectType: 'LanguageTranslationTask',
+        method: 'execute',
+        tenantId,
+        createdAt: { op: '>=', value: since.toISOString() },
```
willgriffin: `SmrtCollection.list()` expects comparison operators in the where key, for example `createdAt >=`, not `{ op, value }` objects. This query will fail or match nothing, and the catch below turns that into `0`, so `translationBudgetPerTenantPerDay` can be silently disabled. Please use the supported where shape and avoid swallowing this particular query failure.

packages/languages/src/models/LanguageOverride.ts
```diff
@@ -0,0 +1,159 @@
...
+      throw new Error('LanguageOverride.template must be a string');
+    }
+
+    this.locale = normalizeLocale(this.locale);
+    // `context` is the conflictColumn-friendly scope: 'app' for nullable tenant,
+    // tenantId otherwise. Mirrors the prompt-override convention.
+    this.context = this.tenantId ?? '__app__';
+
+    const previousIdentity = await this.getPersistedIdentity();
+    const result = await super.save();
```
willgriffin: Identity edits need the same special handling as `PromptOverride`. Because persistence upserts on `(key, locale, context)`, changing `key`, `locale`, or `tenantId` on an existing row makes `super.save()` try to insert the old `id` under a new conflict identity, which can hit the primary-key constraint instead of moving the override. Either disallow identity changes through update/edit APIs or copy the `saveAfterIdentityChange` pattern from prompts before calling `super.save()`.

packages/languages/src/translation-job.ts
```diff
@@ -0,0 +1,401 @@
...
+      typeof parsed.translation === 'string'
+    ) {
+      const cleaned = parsed.translation.trim();
+      return cleaned.length > 0 ? cleaned : null;
+    }
+  } catch {
+    // Fall through — the model may have returned a bare string.
+  }
+
+  const trimmed = String(message).trim();
```
willgriffin, line 446: When `responseFormat: json_object` returns valid JSON with the wrong shape, for example `{"text":"Hola"}`, this falls through and persists the whole JSON blob as the translation. The fallback-to-bare-string path should only run when JSON parsing fails or when the parsed value is itself a string; parsed objects without a string `translation` should return null and fail the job.

packages/languages/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  invalidateResolvedLanguageCache,
+  resolveLanguageString,
+} from './language-resolver.js';
+
+export {
+  LanguageOverride,
+  type LanguageOverrideCtorOptions,
+} from './models/LanguageOverride.js';
+
+export {
```
willgriffin: This static root export defeats the lazy import in `resolveLanguageString`: importing the package root for only `defineLanguageString` or `resolveLanguageString` still evaluates `translation-job.ts` and, through `cli.ts`, pulls in jobs/features/prompts/AI. Vite already warns that the dynamic import cannot split because of these static imports. Consider moving job/CLI APIs to a secondary entry point or lazily importing `enqueueTranslationJob` inside `translateMissing` so normal resolution does not drag the translation worker stack into every consumer bundle.

packages/languages/src/translation-job.ts
```diff
@@ -0,0 +1,401 @@
...
+
+    const prompt = await resolvePrompt(TRANSLATION_PROMPT_KEY, {
+      tenantId: args.tenantId ?? null,
+      variables: {
+        sourceLocale,
+        targetLocale,
+        template: args.sourceTemplate,
+        glossary,
+      },
+    });
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. `execute()` now passes `db: this.options.db` into `resolvePrompt` so app/tenant prompt overrides take effect — that's the whole reason we register the prompt with smrt-prompts in the first place.

packages/languages/src/translation-job.ts
```diff
@@ -0,0 +1,401 @@
...
+    targetLocale,
+    tenantId: options.tenantId ?? null,
+  };
+
+  const job = await jobs.create({
+    queue: 'languages',
+    objectType: 'LanguageTranslationTask',
+    objectId: null,
+    method: 'execute',
+    args: { ...payload, _dedupId: dedupId },
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. `enqueueTranslationJob` now stamps `tenantId: options.tenantId ?? null` on the SmrtJob row so the per-tenant daily budget query (`countTodayTenantTranslations` filters by tenantId) actually counts it.

packages/languages/src/translation-job.ts
```diff
@@ -0,0 +1,401 @@
...
+
+    if (containsObviousMarkupLeak(translated)) {
+      throw new Error(
+        `LanguageTranslationTask: refusing to persist suspicious translation for "${args.key}"`,
+      );
+    }
+
+    const aiModel = (prompt.ai?.model as string | undefined) ?? null;
+    const sourceHash =
+      args.sourceHash || computeSourceHash(args.sourceTemplate);
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. Single `mergedAiConfig` now feeds both `getAI()` and `ai.message()`, so the per-job `args.model` override actually takes effect end-to-end. `ai_model` is persisted from the effective model (`args.model ?? promptAi.model`), not the prompt default.

packages/languages/src/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+  if (!trimmed) return '';
+  // Lowercase language; uppercase region for consistent matching.
+  const [language, ...rest] = trimmed.split('-');
+  if (rest.length === 0) {
+    return language.toLowerCase();
+  }
+  return [
+    language.toLowerCase(),
+    ...rest.map((part) => part.toUpperCase()),
+  ].join('-');
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. `normalizeLocale` docstring now reflects the actual behavior (lowercases language, uppercases everything after) and explicitly notes it is stricter than canonical BCP-47 casing.

packages/languages/src/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/** Render `{var}` placeholders. Missing variables collapse to empty strings. */
+export function renderTemplate(
+  template: string,
+  variables?: LanguageVariables,
+): string {
+  if (!variables || Object.keys(variables).length === 0) {
+    return template;
+  }
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. `renderTemplate` docstring now explains the fast-path: missing variables collapse only when at least one variable is provided; `{}` returns the template unchanged. Added a corresponding clarifying note in the test file.

packages/languages/src/language-resolver.ts
```diff
@@ -0,0 +1,300 @@
...
+ * The actual job machinery lives in `./translation-job.ts` and is loaded
+ * lazily so the resolver does not pull `@happyvertical/ai` into the import
+ * graph for synchronous resolves that never miss.
+ */
+async function tryEnqueueTranslation(args: {
+  key: string;
+  requestedLocale: string;
+  defaultLocale: string;
+  options: ResolveLanguageStringOptions;
+  fallbackTemplate?: string | null;
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. Removed the unused `fallbackTemplate` parameter from `tryEnqueueTranslation` and its call site.

packages/languages/src/language-resolver.ts
```diff
@@ -0,0 +1,300 @@
...
+          resolvedFromLocale: locale,
+        },
+      });
+    }
+  }
+
+  // 3. File-config override.
+  const configTemplate = config.overrides?.[key]?.[locale];
+  if (typeof configTemplate === 'string') {
+    return finalize({
```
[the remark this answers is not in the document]
willgriffin: Fixed in ddcd1a7f. File-config override lookup now normalizes the locale key — `overrides[key]['fr-ca']` matches a `fr-CA` request via a new `lookupNormalizedLocale` helper. New test at `language-resolver.test.ts`: 'honors mixed-case file-config locale keys'.

packages/languages/src/translation-job.ts
willgriffin: Fixed in ddcd1a7f. The default prompt template no longer contains literal braces. Brace-containing examples (placeholder syntax `"{name}"`, JSON shape `{"translation": "Hola, {name}"}`) are now passed in as variables (`placeholderExample`, `responseShapeExample`) so they reach the model verbatim instead of being eaten by `renderPromptTemplate`.

packages/languages/src/translation-job.ts
willgriffin: Fixed in ddcd1a7f. `countTodayTenantTranslations` now uses the supported `'createdAt >=': since.toISOString()` where shape, and lets the call propagate instead of swallowing failures — silent budget bypass would have been a real production hazard.

packages/languages/src/models/LanguageOverride.ts
willgriffin: Fixed in ddcd1a7f. `LanguageOverride.save()` now copies `saveAfterIdentityChange` from PromptOverride: when `(key, locale, tenantId)` changes, it does the delete-then-insert dance in a transaction (or a deferred-delete fallback) instead of letting `super.save()` collide with the PK. New test at `language-resolver.test.ts`: 'moves the row when (key, locale, tenantId) changes'.

packages/languages/src/translation-job.ts
willgriffin: Fixed in ddcd1a7f. `parseTranslationResponse` now treats wrong-shape JSON (e.g. `{"text":"Hola"}`) as an error and returns null. Only the parse-failed branch tries the bare-string fallback. New integration test asserts the wrong-shape response triggers `'returned no usable translation'` and the row is never persisted.

packages/languages/src/index.ts
willgriffin: Fixed in ddcd1a7f. Translation-worker exports moved to a subpath: `@happyvertical/smrt-languages/jobs` (`enqueueTranslationJob`, `LanguageTranslationTask`, `AUTO_TRANSLATE_FEATURE_KEY`, `TRANSLATION_PROMPT_KEY`). Admin CLI helpers moved to `@happyvertical/smrt-languages/cli`. Package root `index.ts` now bundles to 6.77 kB (down from 30.92 kB) and no longer drags `@happyvertical/ai`/jobs/features/prompts into resolver-only consumers; the dynamic import in the resolver still triggers a translation enqueue on miss.

willgriffin:
## Review fixes (ddcd1a7f)

Resolved all 12 review comments. Tests now at 33 passing.

### Real bugs fixed
- **Prompt template literal-brace bug** ([translation-job.ts](packages/languages/src/translation-job.ts)) — placeholders the model needed to see (`"{name}"`, `{"translation": "..."}`) were being eaten by `renderPromptTemplate`. Now passed as variables.
- **Per-tenant daily budget silently disabled** — `countTodayTenantTranslations` used `{ op: '>=', value: ... }` which `SmrtCollection.list` doesn't support, and the failure was swallowed. Switched to `'createdAt >='` and let it propagate.
- **`resolvePrompt` ignored tenant overrides** — wasn't getting the task DB. Now passes `db: this.options.db`.
- **Per-job model override didn't apply to `ai.message()`** — single `mergedAiConfig` now feeds both `getAI()` and `ai.message()`, and `ai_model` records the effective model.
- **`enqueueTranslationJob` didn't stamp `tenantId`** — budget query couldn't see jobs queued outside `withTenant()`. Fixed.
- **Identity-change save would hit PK constraint** — copied `saveAfterIdentityChange` pattern from `PromptOverride`.
- **Wrong-shape JSON falling through to bare-string path** — `{"text":"Hola"}` would have been persisted as the translation. Now treated as an error.

### Other
- File-config override lookup is now case-insensitive (`overrides.fr-ca` matches a `fr-CA` request).
- Removed unused `fallbackTemplate` parameter.
- Docstring corrections on `normalizeLocale` and `renderTemplate`.
- **Subpath exports**: translation-worker stack lives on `@happyvertical/smrt-languages/jobs`, admin CLI on `@happyvertical/smrt-languages/cli`. Package root went from 30.92 kB → 6.77 kB and no longer drags `@happyvertical/ai`, smrt-jobs, smrt-features, or smrt-prompts into resolver-only consumers.

### New tests
- `language-resolver.test.ts` — identity-change save (move across `key` + `locale` + `tenantId` in one save).
- `language-resolver.test.ts` — mixed-case file-config locale keys (lowercase `fr-ca` matches `fr-CA`).
- `integration.test.ts` — wrong-shape JSON rejection (handler throws, no row persisted).

PR #1203, opened 2026-05-09
[codex] Export content editor building blocks

willgriffin:
## Summary
- Export smaller content editor building blocks from `@happyvertical/smrt-content/svelte`
- Add reusable metadata, status, references, image browser, and form helper modules
- Re-export editor assistant/body-format/publish-readiness helpers through the Svelte entrypoint so apps can assemble their own editor layouts without reaching into internals

## Why
Anytown needs to compose the WYSIWYG editor, image browser, metadata/status sections, and right-sidebar assistant in its own page layout while keeping those pieces package-owned and reusable.

## Validation
- `pnpm --dir packages/content check`
- pre-push `format-check`
- pre-push `validate-all-workflows`

### DIFF
--- a/packages/content/src/svelte/components/ContentEditor.svelte
+++ b/packages/content/src/svelte/components/ContentEditor.svelte
@@ -2,11 +2,7 @@
 import type { ImageLike } from '@happyvertical/smrt-images/svelte';
 import { ImageUploader } from '@happyvertical/smrt-images/svelte';
 import { untrack } from 'svelte';
-import {
-  type ContentBodyFormat,
-  extractBodyImages,
-  resolveBodyFormat,
-} from '../../body-format';
+import { extractBodyImages, resolveBodyFormat } from '../../body-format';
 import type {
   ContentEditorAssistantActions,
   ContentEditorAssistantContextChange,
@@ -19,6 +15,13 @@ import type {
   FactEvidenceStatus,
 } from '../../mock-smrt-client';
 import { joinApiUrl, normalizeApiBaseUrl } from '../api';
+import {
+  type ContentEditorFormData,
+  getContentEditorInitialFormData,
+  getContentEditorSavePayload,
+  getContentEditorSnapshot,
+  normalizePublishDate,
+} from '../content-editor-form';
 import ContentAgentChat from './ContentAgentChat.svelte';
 import ContentBodyEditor, {
   type ContentBodyEditorChange,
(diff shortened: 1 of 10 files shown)


willgriffin (commented):
Review of the new content-editor building blocks. Overall the decomposition is reasonable and the goal — letting Anytown compose its own editor layout from package-owned pieces — is sound. A few concerns worth addressing before merge:

**1. `getContentEditorSavePayload` diverges from `ContentEditor.svelte`'s in-component `getSavePayload` (highest-priority issue).**

`packages/content/src/svelte/content-editor-form.ts:97-104` strips both `references` and `assets` from the payload:
```ts
const { references: _references, assets: _assets, ...payload } = data;
```

But `packages/content/src/svelte/components/ContentEditor.svelte:98-105` only strips `references`:
```ts
function getSavePayload(data: any) {
  const { references: _references, ...payload } = data;
  return { ...payload, publish_date: normalizePublishDate(data.publish_date) };
}
```

Result: a consumer assembling a custom editor with the new helper sends a different payload to the API than `ContentEditor` does (no `assets` field). One of these is wrong, and silently. Pick one and fix the other — most likely the new helper's stricter version is correct (you persist IDs, not hydrated objects), in which case `ContentEditor` should be refactored to use the shared helper here.

**2. Duplicated logic instead of refactor.**

`formatDateTimeLocal`, `normalizePublishDate`, the initial form-data shape, and the save payload are now defined in two places (`ContentEditor.svelte` lines 66–105/117–149 and `content-editor-form.ts`). The PR description frames this as extracting reusable helpers, but `ContentEditor.svelte` was not updated to consume them. Two implementations of the same logic will drift (they already have, per #1). Recommend: refactor `ContentEditor.svelte` to import `getContentEditorInitialFormData`, `getContentEditorSavePayload`, and `getContentEditorSnapshot` from the new module, removing the in-component copies.

**3. `ContentEditorFormData` extends `Record<string, any>`.**

`content-editor-form.ts:4` makes the typed fields effectively advisory — any property name with any value type type-checks. If the goal is to give consumers a real contract, drop the index signature. If untyped passthrough is genuinely needed (for STI meta fields, etc.), at least keep the index value as `unknown` rather than `any`.

**4. No tests for any of the new modules.**

Neighbouring components in `packages/content/src/svelte/components/` ship `*.test.ts` (`ContentEditor.test.ts`, `ContentContributionForm.test.ts`, etc.). The four new components and the new helper module add 700+ lines with zero tests. Per `.claude/rules/testing.md`, building blocks like these should at minimum have unit tests for the pure helpers (`getContentEditorSavePayload`, `getContentEditorInitialFormData`, `formatDateTimeLocal` / `normalizePublishDate` round-trips) and rendering smoke tests for each component.

**5. `assets`, `references`, and `factAudit` typed as `any[]` / `Record<string, any>`.**

In `ContentImageBrowser.svelte` and `ContentReferencesPanel.svelte`, asset/reference props are `any[]`, then accessed via long fallback chains:
```ts
asset?.sourceUri || asset?.thumbnailUri || asset?.thumbnailUrl ||
  asset?.deliveryUrl || asset?.url || asset?.src || ''
```
This package owns the `Asset` type — these building blocks should expose it (or at minimum a small `ContentBrowserAsset` interface) rather than the fallback ladder. A consumer assembling a layout has no way to know which fields the component actually expects without reading the source. `factAudit?: Record<string, any> | null` in `ContentReferencesPanel` is similar — `FactAuditStateData` already exists and is used in `ContentEditor.svelte:32`.

**6. `ContentImageBrowser` — `showUploaderInitially` effect is one-way.**

`ContentImageBrowser.svelte:33-37`:
```ts
$effect(() => {
  if (showUploaderInitially) { showUploader = true; }
});
```
Flipping the prop true → false leaves `showUploader = true`. Either honour the prop both directions (`showUploader = showUploaderInitially`) or rename it to clarify it's a one-shot trigger (e.g. `defaultShowUploader`).

**7. `ContentReferencesPanel` reference input is unlabeled.**

`packages/content/src/svelte/components/ContentReferencesPanel.svelte:99-108` uses placeholder text only. Add an `aria-label` ("Add reference by ID or URL") or a `<label>`.

**Minor / non-blocking**
- `ContentImageBrowser.svelte` `` tags lack `loading="lazy"`; matters at scale.
- `{#each ... (asset.id || getAssetLabel(asset))}` and similar for references: when `id` is missing, fallback-to-label keys can collide. Not a bug today (refs always have ids from `referenceIds`) but a footgun if asset shape changes.
- `getContentEditorInitialFormData` accepts `content: any`. With `ContentEditorFormData` already imported here, an `Article | Document | Mirror | Partial<ContentEditorFormData>` union (or just `Partial<ContentEditorFormData>`) would be more useful to consumers.

PR #1204, opened 2026-05-09
feat(content): add feed source imports

willgriffin:
## Summary
- add ContentFeedSource and ContentFeedSourceCollection to model external RSS/Atom feeds
- add feed parser and sync helpers that import feed items into Mirror content
- extend Mirror metadata for source attribution, original URLs, GUIDs, and dedupe keys

## Verification
- pnpm --dir packages/prompts build
- pnpm --dir packages/content check
- pnpm --dir packages/content test -- content-feed-parser.test.ts
- git diff --check HEAD~1..HEAD

## Companion
- Supports https://github.com/anytown/anytown.ai/pull/307.


packages/content/src/content-feed-sync.ts
```diff
@@ -0,0 +1,305 @@
...
+}
+
+async function findExistingMirror(
+  contents: QueryableCollection,
+  source: ContentFeedSource,
+  item: ParsedContentFeedItem,
+  dedupeKey: string,
+): Promise<{ id: string; status: string | null } | null> {
+  const tenantWhere = source.tenantId ? 'tenant_id = ?' : 'tenant_id IS NULL';
+  const params: unknown[] = ['Mirror'];
```
willgriffin: **STI discriminator must be the qualified name, not the bare class name.**

This hard-codes `'Mirror'` as the `_meta_type` discriminator (and again at L180 in the INSERT). The ORM writes the qualified name `@happyvertical/smrt-content:Mirror` via `getResolvedQualifiedName()` ([packages/core/src/object.ts](https://github.com/happyvertical/smrt/blob/3b38857c5a7765a14250773a0285dfe020c9beab/packages/core/src/object.ts#L917-L922)), and there is an explicit regression test pinning this format in [issue-1170-sti-legacy-discriminator-save.test.ts](https://github.com/happyvertical/smrt/blob/3b38857c5a7765a14250773a0285dfe020c9beab/packages/core/src/__tests__/issue-1170-sti-legacy-discriminator-save.test.ts#L52-L60).

Per [CLAUDE.md](https://github.com/happyvertical/smrt/blob/3b38857c5a7765a14250773a0285dfe020c9beab/CLAUDE.md#L86-L90):

> **STI discriminator**: qualified names — `@happyvertical/smrt-content:Article`

**Effect:** `findExistingMirror`'s `WHERE _meta_type = 'Mirror'` won't match Mirror rows the ORM has previously written (which use the qualified name), so `upsertMirrorItem` will INSERT a duplicate row on every sync. Conversely, the rows this code inserts with `_meta_type = 'Mirror'` won't be loaded by `Mirrors.list()` or any other STI-aware query.

Replace `'Mirror'` with `'@happyvertical/smrt-content:Mirror'` at L71 and L180. Ideally derive it from the registry rather than hard-coding the string.

packages/content/src/content-feed-sync.ts
```diff
@@ -0,0 +1,305 @@
...
+  };
+  if (source.etag) headers['If-None-Match'] = source.etag;
+  if (source.lastModified) headers['If-Modified-Since'] = source.lastModified;
+
+  const now = options.now?.() ?? new Date();
+  source.markFetchStarted(now);
+  await source.save();
+
+  try {
+    const response = await fetchImpl(source.feedUrl, { headers });
```
willgriffin: **SSRF / unvalidated outbound fetch.**

`source.feedUrl` originates from user input via the auto-generated `POST /api/v1/contentfeedsources` route, and is fetched here with no protocol allowlist, no host validation, no timeout, and no response-size cap. The neighboring `feedUrl` field declaration in [content-feed-source.ts](https://github.com/happyvertical/smrt/blob/3b38857c5a7765a14250773a0285dfe020c9beab/packages/content/src/content-feed-source.ts) is just `text({ required: true, indexed: true })` — there's no validation hook upstream either.

A user who can create a `ContentFeedSource` can point `feedUrl` at loopback / RFC1918 / link-local addresses (e.g. `http://169.254.169.254/latest/meta-data/`, `http://127.0.0.1:6379/`). The response body is then parsed and either stored as Mirror content rows (data exfil) or surfaced via `lastError` on the row the attacker owns. With no fetch timeout (Node's global `fetch` has none) and no cap on `response.text()`, a hostile or slow target is also a straightforward DoS.

Suggested mitigations:
- Validate `feedUrl` scheme is `http:`/`https:` (on the model field or before the fetch).
- Resolve the host and reject loopback / private / link-local addresses, or use an allowlist of trusted feed hosts.
- Wrap the fetch with `AbortSignal.timeout(...)` and cap the body length read from `response.text()`.

willgriffin (commented):
## Code review

Three issues worth addressing before this lands. Two are real bugs (one security, one logic) and one is a CLAUDE.md violation that will silently break STI lookups against ORM-written rows.

Not flagged: I checked the `Mirror` STI fields (no `@meta()`) and the `metadata: Record<string, any>` field on `ContentFeedSource` — both are consistent with the prevailing patterns in the codebase (e.g. `Email` STI child uses plain fields too; `metadata: Record<string, any>` is the documented carve-out in `.claude/rules/smrt-patterns.md` and matches `Content`, `Account`, `Secret`, etc.).

packages/content/src/content-feed-sync.ts
```diff
@@ -0,0 +1,305 @@
...
+    );
+    if (result[0]) return result[0] as { id: string; status: string | null };
+  }
+
+  const result = await contents.query(
+    `SELECT id, status
+       FROM contents
+      WHERE _meta_type = ?
+        AND ${tenantWhere}
+        AND (dedupe_key = ? OR external_url = ? OR original_url = ?)
```
willgriffin: **Cross-source clobber: fallback URL match isn't scoped by `feed_source_id`.**

This fallback only filters by `_meta_type` and tenant. When two `ContentFeedSource`s in the same tenant share an item URL — very common with syndicated / republished / aggregator content — Source B's sync misses the GUID-scoped query and falls through to this OR-on-URL match, where it finds Source A's existing Mirror row.

`upsertMirrorItem` then [UPDATEs](https://github.com/happyvertical/smrt/blob/3b38857c5a7765a14250773a0285dfe020c9beab/packages/content/src/content-feed-sync.ts#L120-L148) overwriting `feed_source_id`, `source_guid`, `source_name`, `source_homepage_url`, and `dedupe_key` with Source B's values. On Source A's next sync the same fallback path matches and rewrites them back. The two sources perpetually clobber each other's row instead of each owning its own Mirror — which also means duplicate-imported attribution and a churning `dedupe_key` for any URL that overlaps.

Add `AND feed_source_id = ?` to the fallback query, or drop the URL predicates and match only on `dedupe_key` (which is already namespaced per-source via `createDedupeKey`).

PR #1224, opened 2026-05-14
[codex] Expose content editor primitives

willgriffin:
## Summary

This lands the source-side fix for the content editor exports that Anytown PR 312 needed from `@happyvertical/smrt-content/svelte`.

- adds exported composable editor primitives: `ContentTitleField`, `ContentReviewStatusTray`, `createContentEditorState`, and image selection/source helpers
- preserves the existing `ContentEditor` and `GovernedContentEditor` exports for compatibility
- adds regression coverage for shared editor state, undo behavior, asset tracking, and preview source resolution
- documents the primitives in the content package README/CLAUDE notes

## Release note

`@happyvertical/smrt-content@0.24.18` was already published from a side branch before these source changes landed on `main`. This PR includes an explicit minor changeset so the next official release uses a fresh version instead of attempting to reuse that package version. The existing publish workflow already corrects 0.x fixed-group major output back to a 0.x minor release.

## Validation

- `pnpm --filter @happyvertical/smrt-prompts build`
- `pnpm --dir packages/content test`
- `pnpm --dir packages/content check`
- `pnpm --dir packages/content build && pnpm --dir packages/content verify:pack`
- `pnpm exec biome check packages/content/CLAUDE.md packages/content/README.md packages/content/src/svelte/index.ts packages/content/src/svelte/content-editor-primitives.test.ts packages/content/src/svelte/content-editor-media.ts packages/content/src/svelte/content-editor-state.svelte.ts packages/content/src/svelte/components/ContentReviewStatusTray.svelte packages/content/src/svelte/components/ContentTitleField.svelte`

The first `packages/content check` failed before `@happyvertical/smrt-prompts` was built in the fresh worktree; after building that workspace dependency, it passed with 0 Svelte errors.

### DIFF
--- a/packages/content/src/svelte/components/ContentReviewStatusTray.svelte
+++ b/packages/content/src/svelte/components/ContentReviewStatusTray.svelte
@@ -1,0 +1,152 @@
+<script lang="ts">
+export type ContentReviewStatusTone =
+  | 'neutral'
+  | 'danger'
+  | 'warning'
+  | 'success';
+export type ContentReviewStatusIcon = 'content' | 'facts' | 'safety' | 'custom';
+
+export interface ContentReviewStatusTrayItem {
+  id: string;
+  icon: ContentReviewStatusIcon;
+  label: string;
+  tone: ContentReviewStatusTone;
+  status: string;
+}
+
+export interface Props {
+  items?: ContentReviewStatusTrayItem[];
+  activeId?: string | null;
+  open?: boolean;
+  label?: string;
+  onSelect?: (item: ContentReviewStatusTrayItem) => void;
+}
+
+let {
+  items = [],
+  activeId = null,
+  open = false,
+  label = 'Review status',
+  onSelect = undefined,
+}: Props = $props();
+</script>
+
+{#snippet trayIcon(icon: ContentReviewStatusIcon)}
+  {#if icon === 'content'}
+    <svg viewBox="0 0 24 24" aria-hidden="true">
+      <path d="M9 11l2 2 4-4"></path>
+      <path d="M8 5h8"></path>
+      <path d="M8 19h8"></path>
+      <path d="M5 7h14"></path>
+      <path d="M5 17h14"></path>
+    </svg>
+  {:else if icon === 'facts'}
+    <svg viewBox="0 0 24 24" aria-hidden="true">
+      <path d="M14 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V8Z"></path>
+      <path d="M14 2v6h6"></path>
+      <path d="M9 15h6"></path>
+      <path d="M9 11h2"></path>
+    </svg>
+  {:else if icon === 'safety'}
+    <svg viewBox="0 0 24 24" aria-hidden="true">
+      <path d="M12 22s8-4 8-10V5l-8-3-8 3v7c0 6 8 10 8 10Z"></path>
+      <path d="m9 12 2 2 4-4"></path>
+    </svg>
+  {:else}
+    <svg viewBox="0 0 24 24" aria-hidden="true">
+      <path d="m12 3 1.8 4.2L18 9l-4.2 1.8L12 15l-1.8-4.2L6 9l4.2-1.8Z"></path>
+      <path d="m5 15 .9 2.1L8 18l-2.1.9L5 21l-.9-2.1L2 18l2.1-.9Z"></path>
+      <path d="m19 14 .9 2.1L22 17l-2.1.9L19 20l-.9-2.1L16 17l2.1-.9Z"></path>
+    </svg>
+  {/if}
+{/snippet}
+
+{#if items.length > 0}
+  <div class="content-review-status-tray" role="tablist" aria-label={label}>
+    {#each items as item (item.id)}
+      <button
+        type="button"
+        role="tab"
+        aria-selected={activeId === item.id && open}
+        class={`tray-button tray-button--${item.tone}`}
+        class:active={activeId === item.id && open}
+        title={`${item.label}: ${item.status}`}
+        onclick={() => onSelect?.(item)}
+      >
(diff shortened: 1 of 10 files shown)


packages/content/src/svelte/content-editor-state.svelte.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    for (const key of Object.keys(fields)) {
+      oldValues[key] = String(this._form[key] ?? '');
+      this._form[key] = fields[key];
+    }
+
+    this._fieldUndoStack = [...this._fieldUndoStack, oldValues];
+    this._lastAppliedFields = Object.keys(fields);
+    this._showUndoBanner = true;
+  }
```
willgriffin: Two related concerns now that this is a public primitive:

1. **Lossy undo for non-string fields.** `oldValues[key] = String(this._form[key] ?? '')` stringifies whatever was there. If a caller passes a key like `tags` or `thumbnailAssetId`, undo restores them as `'a,b,c'` / `''` instead of the original array / `null`. The existing assistant flow gates this via `sanitizeContentEditorAssistantFieldUpdates` (text + enum keys only), but nothing in this class enforces that.
2. **Type mismatch on write.** `fields[key]` is `string | number | boolean | null | undefined`, but `_form` fields are typed (e.g. `title: string`, `thumbnailAssetId: string | null`). Writing the raw value can drop the form into a state that violates `ContentEditorFormData`.

Options: (a) narrow the parameter type to `Record<string, string>` to match what the assistant actually sends and what `oldValues` assumes, or (b) store `oldValues` as `Record<string, unknown>` and coerce/validate new values before writing. Either way I'd add a test that exercises a non-string field through this path so the contract is locked in.

packages/content/src/svelte/content-editor-media.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  if (typeof selected === 'string') {
+    const parsedUrl = new URL(selected);
+    return createContentEditorImageRecord(apiBaseUrl, {
+      name: parsedUrl.pathname.split('/').pop() || 'External Image',
+      sourceUri: selected,
+      mimeType: 'image/jpeg',
+    });
+  }
```
willgriffin, line 112: Two issues in the string branch:

1. `new URL(selected)` throws on any non-absolute / malformed URL — that rejects the promise with a `TypeError` rather than returning `null` like the other branches do. Wrap in `try`/`catch` or pre-validate with `URL.canParse(selected)` and return `null` on failure.
2. `mimeType: 'image/jpeg'` is hardcoded regardless of extension, so a `.png` / `.gif` / `.webp` URL gets recorded as JPEG. Derive from the pathname extension (or pass `application/octet-stream` if unknown) — anything is better than silently mislabeling.

packages/content/src/svelte/components/ContentReviewStatusTray.svelte
```diff
@@ -0,0 +1,152 @@
...
+        class={`tray-button tray-button--${item.tone}`}
+        class:active={activeId === item.id && open}
+        title={`${item.label}: ${item.status}`}
+        onclick={() => onSelect?.(item)}
+      >
+        {@render trayIcon(item.icon)}
+        <span class="tray-button__indicator" aria-hidden="true"></span>
+        <span class="sr-only">{item.label}: {item.status}</span>
+      </button>
+    {/each}
```
willgriffin, line 79: The `role="tablist"` + `role="tab"` pairing implies the [WAI-ARIA tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), which expects each tab to have `aria-controls` pointing at a `role="tabpanel"` and arrow-key roving focus through the tablist. Neither is wired up here — the buttons just call `onSelect`, and the actual panel lives outside this component.

If these are really tabs, add `aria-controls` (caller-supplied) and arrow-key handling. If they're more like a toggle group revealing a panel elsewhere, `role="button"` with `aria-pressed={activeId === item.id && open}` describes the behavior more honestly and avoids screen-reader confusion.

willgriffin (commented):
Overall this is a clean extraction — public surface is well-scoped and the regression tests are a nice safety net. A few concerns below, mostly around the new public `applyFieldUpdates` contract being looser than the assistant flow it was originally written for, and a couple of robustness issues in the new media helpers.

**Summary**
- Composable primitives (`ContentTitleField`, `ContentReviewStatusTray`, `createContentEditorState`, media helpers) are exported cleanly; existing `ContentEditor`/`GovernedContentEditor` remain untouched.
- Test coverage hits the happy paths for form init, undo, asset add/remove, and preview source resolution — good.
- README/CLAUDE docs updated alongside the exports.

**Notable issues**
- `applyFieldUpdates` lossily stringifies the old value and writes raw assistant-typed values back into the form (see inline). Fine for the existing sanitized assistant flow (which only emits `Record<string, string>`), but the class is now a public primitive — callers passing non-string values or non-text/enum keys can corrupt array/null fields, and undo will then restore them as strings.
- `resolveContentEditorImageSelection` calls `new URL(selected)` unguarded — a non-URL string throws synchronously inside an async function (rejects the promise with `TypeError`), which is a worse failure mode than the documented `null` return. Mime type for URL strings is also hardcoded to `image/jpeg`.
- `ContentReviewStatusTray` advertises `role="tablist"`/`role="tab"` but doesn't pair the buttons with `aria-controls`/`tabpanel` or implement roving focus — under the WAI-ARIA tabs pattern this is incomplete. If the intent is a selectable button group, `role="button"` + `aria-pressed` is a better fit.

**Nits**
- Mutation style inside `ContentEditorState` is inconsistent — `update()` replaces the whole `_form` object, while `addAsset`/`removeAsset` mutate fields in place. Either works with Svelte 5 `$state` proxies, but picking one keeps reactivity reasoning simpler.
- `font-weight: 720` in `ContentTitleField` is only meaningful for variable fonts; most static font stacks will round to 700 — worth a comment if that's intentional.

Issue #1165, opened 2026-04-28
Tenant upsert generates ON CONFLICT(slug, context, _meta_type) but only (slug, context) has a unique constraint — STI tables fail to save


willgriffin:
## Summary

Saving a `Tenant` (or any STI-discriminated subclass — `Network`, `Publication`, etc.) via the standard `collection.create()` / `instance.save()` flow fails on PostgreSQL with:

```
DatabaseError: Database query failed: UPSERT INTO tenants
Cause: there is no unique or exclusion constraint matching the
ON CONFLICT specification, code=42P10, severity=ERROR
```

The generated SQL targets `ON CONFLICT(slug, context, _meta_type)`, but the live schema for `tenants` only carries unique constraints on `(slug, context)` — the trailing `_meta_type` column makes it a regular non-unique index.

## Repro

1. Provision a fresh anytown.ai dashboard install (\`@happyvertical/smrt-core\` 0.22.14, \`@happyvertical/sql\` 0.71.32, PostgreSQL 17).
2. Run \`pnpm smrt db:migrate\` to materialise the schema from the SMRT manifest.
3. Sign in as a super admin and open \`/super/tenants\`.
4. Click **New Tenant**, enter \`Doctor Test\` / \`doctor-test-1\`, type **Network**, no parent.
5. Submit.

The action returns \`{ success: false, error: 'Failed to create tenant' }\` and the dev log shows the trace above.

The same error reproduces calling \`networkCollection.create({ name, slug, status: 'active' })\` directly.

## Expected vs actual

**Expected:** the upsert succeeds — either by emitting \`ON CONFLICT(slug, context)\` (which matches the unique index that exists), or by ensuring the migration emits a unique constraint on \`(slug, context, _meta_type)\` to match what the upsert plan generates.

**Actual:** every save against an STI table is poisoned. We can't create new tenants through the UI.

## Indexes on \`tenants\` (anytown.ai)

\`\`\`
tenants_id_idx                      btree (id)
tenants_meta_type_idx               btree (_meta_type)
tenants_pkey         UNIQUE         btree (id)
tenants_slug_context_idx     UNIQUE btree (slug, context)
tenants_slug_context_key     UNIQUE btree (slug, context)
tenants_slug_context_meta_type_idx  btree (slug, context, _meta_type)        ← non-unique, no good for ON CONFLICT
\`\`\`

[...]

PR #1275, opened 2026-05-20
Make generated schema migrations atomic

willgriffin:
## Summary
- run generated smrt db:migrate schema changes inside a single transaction instead of committing each generated change independently
- roll back table creation and generated column/index migrations on the first failure, then fail loudly
- add MigrationTracker.applyAll atomic support and preserve the original migration error when Postgres aborts the surrounding transaction

## Why
Anytown dev showed the failure mode: a generated migration failed partway through, but earlier table/column changes remained committed. Flux then blocked the app rollout while the database had partial schema side effects.

## Verification
- pnpm exec biome check packages/core/src/migrations/tracker.ts packages/core/src/migrations/types.ts packages/core/src/migrations/__tests__/tracker.test.ts packages/cli/src/commands/utilities.ts packages/cli/src/commands/__tests__/db-migrate-atomic.test.ts
- pnpm --filter @happyvertical/smrt-core exec vitest run src/migrations/__tests__/tracker.test.ts
- pnpm --filter @happyvertical/smrt-cli exec vitest run src/commands/__tests__/utilities.test.ts src/commands/__tests__/db-migrate-atomic.test.ts
- pnpm --filter @happyvertical/smrt-core typecheck
- pnpm --filter @happyvertical/smrt-cli typecheck
- pnpm --filter @happyvertical/smrt-core build
- pnpm --filter @happyvertical/smrt-cli build
- Postgres smoke: failed atomic batch left no first table behind on local Postgres

### DIFF
--- a/packages/cli/src/commands/utilities.ts
+++ b/packages/cli/src/commands/utilities.ts
@@ -10,6 +10,7 @@ import { createRequire } from 'node:module';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import {
+  tableExists as coreTableExists,
   isQualifiedName,
   ObjectRegistry,
   SchemaComparer,
@@ -83,10 +84,6 @@ function formatSchemaCommandFailureHeader(
   return fallback;
 }
 
-function getRows(result: unknown): any[] {
-  return Array.isArray(result) ? result : ((result as any)?.rows ?? []);
-}
-
 function createAtomicSchemaMigrationDb(tx: any, dbType: string): any {
   const migrationDb: any = {
     ...tx,
@@ -96,29 +93,8 @@ function createAtomicSchemaMigrationDb(tx: any, dbType: string): any {
   migrationDb.transaction = async <T>(callback: (txDb: any) => Promise<T>) =>
     callback(migrationDb);
 
-  migrationDb.tableExists = async (tableName: string): Promise<boolean> => {
-    if (dbType === 'postgres') {
-      const result = await migrationDb.query(
-        `SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = ? AND table_schema = 'public') as exists`,
-        tableName,
-      );
-      return Boolean(getRows(result)[0]?.exists);
-    }
-
-    if (dbType === 'duckdb') {
-      const result = await migrationDb.query(
-        `SELECT table_name FROM information_schema.tables WHERE table_name = ?`,
-        tableName,
-      );
-      return getRows(result).length > 0;
-    }
-
-    const result = await migrationDb.query(
-      `SELECT name FROM sqlite_master WHERE type = 'table' AND name = ?`,
-      tableName,
-    );
-    return getRows(result).length > 0;
-  };
+  migrationDb.tableExists = (tableName: string): Promise<boolean> =>
+    coreTableExists(migrationDb, tableName, dbType);
 
   return migrationDb;
 }
@@ -1565,7 +1541,7 @@ export default testManifest;
                   };
 
                   const result = await batchTracker.apply(migrationDef, {
-                    postgresSafe: options['postgres-safe'] ?? false,
+                    postgresSafe: false,
                     force: options.force ?? false,
                   });
 
@@ -1645,7 +1621,7 @@ export default testManifest;
 
                   // Apply migration - tracker handles checksum validation and idempotency
                   const result = await batchTracker.apply(migrationDef, {
-                    postgresSafe: options['postgres-safe'] ?? false,
+                    postgresSafe: false,
                     force: options.force ?? false,
                     // The diff was computed from the live schema moments earlier, so
                     // missing columns/indexes must be repaired even if a previous
--- a/packages/cli/src/commands/utilities.ts
+++ b/packages/cli/src/commands/utilities.ts
@@ -83,6 +83,46 @@ function formatSchemaCommandFailureHeader(
   return fallback;
 }
 
+function getRows(result: unknown): any[] {
+  return Array.isArray(result) ? result : ((result as any)?.rows ?? []);
+}
+
+function createAtomicSchemaMigrationDb(tx: any, dbType: string): any {
+  const migrationDb: any = {
+    ...tx,
+    syncSchema: undefined,
+  };
+
+  migrationDb.transaction = async <T>(callback: (txDb: any) => Promise<T>) =>
+    callback(migrationDb);
+
+  migrationDb.tableExists = async (tableName: string): Promise<boolean> => {
+    if (dbType === 'postgres') {
+      const result = await migrationDb.query(
+        `SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = ? AND table_schema = 'public') as exists`,
+        tableName,
+      );
+      return Boolean(getRows(result)[0]?.exists);
+    }
+
+    if (dbType === 'duckdb') {
+      const result = await migrationDb.query(
+        `SELECT table_name FROM information_schema.tables WHERE table_name = ?`,
+        tableName,
+      );
+      return getRows(result).length > 0;
+    }
+
+    const result = await migrationDb.query(
+      `SELECT name FROM sqlite_master WHERE type = 'table' AND name = ?`,
+      tableName,
+    );
+    return getRows(result).length > 0;
+  };
+
+  return migrationDb;
+}
+
 /**
  * Index definition type for migration comparison
  */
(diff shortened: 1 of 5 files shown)


willgriffin (commented):
Reviewed the atomic-batch approach end-to-end. The core idea is right and the bug it fixes is real (committed-then-aborted partial schemas blocking Flux rollouts in anytown). The follow-up commit `13995269e` already addressed the custom `tableExists` shim and made the in-batch `postgresSafe: false` explicit — those concerns are resolved.

Remaining feedback below. Most of it is about observability and the fact that `MigrationTracker.applyAll({ atomic: true })` is shipped but never called from the CLI; the CLI hand-rolls equivalent logic instead.

packages/cli/src/commands/utilities.ts
```diff
@@ -1530,102 +1488,178 @@ export default testManifest;
...
+                  });
+
+                  if (!result.success) {
+                    throw result.error || new Error(`${migration.type} failed`);
+                  }
+
+                  // Check if this was already applied (execution_time_ms = 0 means skipped)
+                  if (result.execution_time_ms === 0) {
+                    if (options.verbose) {
+                      logLines.push(
```
willgriffin: **The CLI doesn't use the new `applyAll({ atomic: true })`.** It hand-rolls its own `db.transaction(...)` + per-step `batchTracker.apply()` loop here, which is ~130 lines that mirror what the tracker now does in `applyAll`'s atomic branch (tracker.ts:482-546). You ship the atomic API and a unit test for it (tracker.test.ts), but production goes through this duplicate.

Two concrete risks:
1. Behavior drift — e.g. `applyAll`'s atomic branch rejects `CONCURRENTLY` DDL up front via `findConcurrentIndexStatement`, but the CLI path doesn't, so a `--postgres-safe` rewrite that somehow leaks `CONCURRENTLY` into a generated migration would fail inside the transaction here instead of with the nicer pre-flight error.
2. The tracker test asserts atomic-rollback semantics on a path no consumer exercises. The CLI test (`db-migrate-atomic.test.ts`) mocks `MigrationTracker` entirely, so the real cross-piece integration is unverified.

If the CLI called `batchTracker.applyAll(allDefs, { atomic: true, postgresSafe: false, force: ..., reconcile: true })` after building one combined list of `MigrationDefinition`s (create_table_* first, then the partitioned column/index ops), the loop body collapses to just building defs + reading results.

packages/cli/src/commands/utilities.ts
```diff
@@ -1530,102 +1488,178 @@ export default testManifest;
...
+                  const migrationDef = {
+                    id: migrationName,
+                    description: `Create table ${schema.tableName}`,
+                    version: '1.0.0',
+                    up: [upSql],
+                    down: [`DROP TABLE IF EXISTS "${schema.tableName}"`],
+                  };
+
+                  const result = await batchTracker.apply(migrationDef, {
+                    postgresSafe: false,
```
willgriffin: **`--postgres-safe` is now a silent no-op for `db:migrate` whenever there are schema changes.** The flag's documented purpose is to use `CREATE INDEX CONCURRENTLY` / `DROP INDEX CONCURRENTLY` to avoid `ACCESS EXCLUSIVE` lock contention on production tables (#1165 context). Hardcoding `postgresSafe: false` here and at 1624 means a user running `smrt db:migrate --postgres-safe` against a busy Postgres still takes the table-level lock for the duration of the batch — exactly the failure mode `--postgres-safe` exists to avoid.

For a small dev migration this is fine; for production with large tables + index changes it's a regression vs. the pre-PR behavior, just without the partial-schema bug.

Options, in order of effort:
- (cheap) When `options['postgres-safe']` is set and `migrations` contains any `add_index`/`drop_index`, log a one-line warning that index ops will run inside the atomic transaction and won't use `CONCURRENTLY`. Keeps the user-visible promise of the flag honest.
- (medium) Partition: column/table changes in the atomic batch, then index changes in a separate non-atomic pass under `--postgres-safe` so they can use CONCURRENTLY. Index ops are independently reversible and don't need to share a transaction with the column adds. The data-integrity argument for atomicity here is weak — an index missing after a partial run is recoverable without manual cleanup, unlike a column.
- (skip) If neither is acceptable today, at minimum document the trade-off in the `--postgres-safe` help text.

packages/cli/src/commands/utilities.ts
```diff
@@ -1530,102 +1488,178 @@ export default testManifest;
...
+                console.log(line);
               }
+              successCount += batchSuccessCount;
+              skippedCount += batchSkippedCount;
             } catch (error) {
               errorCount++;
               const errorMsg =
                 error instanceof Error ? error.message : String(error);
-              console.error(`  ✗ ${migration.type} failed: ${errorMsg}`);
+              console.error(`  ✗ atomic schema migration failed: ${errorMsg}`);
```
willgriffin: **This error doesn't tell the user which migration failed within the batch.** The throw at 1633 (and 1574 for table creation) is `result.error || new Error(\`${migration.type} failed\`)`, which surfaces here as e.g. `✗ atomic schema migration failed: column "foo" already exists`. The user has no idea whether it was the second column add, the fifth, or one of the index creates.

Compare `applyAll`'s own rollback message: `Rolled back because migration ${error.failed.name} failed` (tracker.ts:537-539) — it includes the migration name.

Fix at the throw sites: `throw new Error(\`${migrationName}: ${result.error?.message ?? migration.type + ' failed'}\`, { cause: result.error })`. Cheap and makes the diagnostic actually actionable for the Anytown-style scenario this PR exists to fix.

packages/cli/src/commands/utilities.ts
```diff
@@ -1530,102 +1488,178 @@ export default testManifest;
...
+                      `  ✓ ${actionDesc} (${shortChecksum(result.checksum)})`,
+                    );
+                    batchSuccessCount++;
+                  }
+                }
+              });
+
+              for (const line of logLines) {
+                console.log(line);
               }
```
willgriffin: **Minor UX: buffered logging hides progress during long batches.** `logLines` accumulates and only prints after `db.transaction(...)` resolves. A multi-minute migration (e.g. several `CREATE INDEX` on a million-row table inside the atomic batch) shows nothing between `🔨 Applying N schema change(s) atomically...` and the final commit — easy to mistake for a hang during a Flux rollout.

Since each `logLines.push(\`  ✓ ...\`)` only runs *after* the corresponding `batchTracker.apply()` returned success in-tx, you could `console.log` it inline instead of buffering. If the batch then rolls back, the user sees the success lines followed by the failure + rollback notice — which is more honest than silence, because those statements did succeed and were then rolled back, and that's exactly what the user needs to understand to debug. (Just adjust the rollback banner copy slightly to make it explicit, e.g. `Rolled back the steps shown above`.)

packages/core/src/migrations/tracker.ts
```diff
@@ -410,14 +441,24 @@ export class MigrationTracker {
...
+          `UPDATE _smrt_schema_migrations
+           SET status = 'failed', error_message = ?
+           WHERE id = ?`,
+          migrationError.message,
+          id,
+        );
+      } catch {
+        // Preserve the migration failure; status persistence is secondary and
+        // may be impossible until the surrounding transaction rolls back.
+      }
```
willgriffin, line 468: **Bare `catch {}` silently swallows every error — not just the Postgres-transaction-abort case the comment describes.** The comment correctly explains why this is needed *inside* an atomic batch on Postgres (the `UPDATE` will fail with `current transaction is aborted, commands ignored until end of transaction block`), but the catch fires for any failure of this UPDATE — including DB connectivity drops, lock timeouts on the migrations table itself, etc. — for *every* `apply()` call, including non-atomic ones.

Pre-PR these would have propagated and the operator would see them. Post-PR they're invisible: caller gets a `MigrationResult` with the original `migrationError` and never knows that the `failed`-status row didn't actually get written. A subsequent `smrt db:status` would then show the row stuck at `running` (or missing entirely if it was the INSERT path), which is a confusing state to debug.

Two low-cost options:
- Narrow the catch to known transaction-abort error codes (Postgres `25P02` / `InFailedSqlTransaction`) and rethrow others.
- Or at minimum `console.error('failed to persist failed-status row for migration X: ...')` in the catch so it's visible in CI/operator logs.

packages/core/src/migrations/tracker.ts
```diff
@@ -55,6 +55,37 @@ const DEFAULT_OPTIONS = {
...
+  constructor(
+    public readonly results: MigrationResult[],
+    public readonly failed: MigrationResult,
+  ) {
+    super(`Atomic migration batch failed at ${failed.name}`);
+  }
+}
+
+const CONCURRENT_INDEX_STATEMENT_RE =
+  /(?:CREATE\s+(?:UNIQUE\s+)?INDEX|DROP\s+INDEX)\s+CONCURRENTLY/i;
```
willgriffin: Nit: `REINDEX CONCURRENTLY` is also forbidden inside a transaction block in Postgres and isn't matched by this regex. Generated migrations don't emit `REINDEX`, so this is fine for the current code path, but if hand-written migrations ever flow through `applyAll({ atomic: true })` it'd be missed. Worth a comment noting the regex is scoped to what the generator emits, or extending it to include `REINDEX\s+(?:.+?)CONCURRENTLY`.

PR #1324, opened 2026-06-02
[codex] Fix test DB STI collection schemas

willgriffin:
## Summary
- Resolve inherited `SmrtCollection` registrations to their item/STI base schema before test DB table creation.
- Skip inherited collection classes in registry schema aggregation so collection stubs cannot create shared STI tables.
- Make manifest `includeObjects` filters map collection subclasses to inherited item schemas while preserving package-scoped qualified filters.

## Root Cause
Manifest collection subclasses such as `EmailCollection` extend a package collection base rather than `SmrtCollection` directly. Test schema prep treated those subclasses as model schemas, so a collection stub sharing an STI table could create the table before the STI base and omit `_meta_type` / `_meta_data`.

During local review, I also found the first include-filter fix widened qualified includes to same-named classes from other packages; this PR keeps qualified filters package-specific and covers that with a regression test.

## Validation
- `pnpm build`
- `pnpm --filter @happyvertical/smrt-core run build`
- `pnpm --filter @happyvertical/smrt-core run typecheck`
- `pnpm --filter @happyvertical/smrt-vitest run typecheck`
- `pnpm exec biome check packages/core/src/testing/database.ts packages/core/src/registry/schema-builder.ts packages/core/src/__tests__/test-database-manifest-schema.test.ts packages/vitest/src/test-db.ts packages/vitest/src/__tests__/test-db-from-manifest.test.ts`
- `(cd packages/core && pnpm exec vitest run src/__tests__/test-database-manifest-schema.test.ts)`
- `(cd packages/vitest && pnpm exec vitest run src/__tests__/test-db-from-manifest.test.ts)`
- `git diff --check`

### DIFF
--- a/packages/core/src/registry/schema-builder.ts
+++ b/packages/core/src/registry/schema-builder.ts
@@ -13,69 +13,24 @@ import type {
   SQLDataType,
 } from '../schema/types.js';
 import { classnameToTablename, toSnakeCase } from '../utils';
+import {
+  type CollectionRegistrationLookup,
+  isCollectionRegistration,
+} from './collection-resolution';
 import { findClass } from './name-resolver';
 import { getClasses, getCollectionTableNames } from './shared-state';
 
 type ForeignKeyAction = NonNullable<
   NonNullable<ColumnDefinition['foreignKey']>['onDelete']
 >;
-type RegisteredSchemaClass = NonNullable<ReturnType<typeof findClass>>;
 
-function isSmrtCollectionExtendsName(extendsName: string | undefined): boolean {
-  return (
-    extendsName === 'SmrtCollection' ||
-    extendsName?.endsWith(':SmrtCollection') === true
-  );
-}
-
-function resolveRelatedRegistration(
-  className: string,
-  origin: RegisteredSchemaClass,
-): RegisteredSchemaClass | undefined {
-  if (className.includes(':')) {
-    return findClass(className);
-  }
-
-  if (origin.packageName) {
-    return (
-      ObjectRegistry.getClassInPackage(origin.packageName, className) ||
-      findClass(className)
-    );
-  }
-
-  return findClass(className);
-}
-
-function getCollectionAncestorRegistrations(
-  className: string,
-  registered: RegisteredSchemaClass,
-): RegisteredSchemaClass[] {
-  const lookupName = registered.qualifiedName || registered.name || className;
-  const chain = ObjectRegistry.getInheritanceChain(lookupName);
-  const registrations: RegisteredSchemaClass[] = [];
-
-  for (const ancestorName of chain) {
-    const ancestor = resolveRelatedRegistration(ancestorName, registered);
-    if (ancestor) {
-      registrations.push(ancestor);
-    }
-  }
-
-  return registrations;
-}
-
-function isCollectionRegistration(
-  className: string,
-  registered: RegisteredSchemaClass,
-): boolean {
-  if (isSmrtCollectionExtendsName(registered.extends)) {
-    return true;
-  }
-
-  return getCollectionAncestorRegistrations(className, registered).some(
-    (ancestor) => isSmrtCollectionExtendsName(ancestor.extends),
-  );
-}
+const collectionRegistrationLookup: CollectionRegistrationLookup = {
+  findClass,
+  findClassInPackage: (packageName, className) =>
+    ObjectRegistry.getClassInPackage(packageName, className),
+  getInheritanceChain: (className) =>
+    ObjectRegistry.getInheritanceChain(className),
+};
 
 function applyDecoratorSqlTypeOverrides(
   className: string,
@@ -220,7 +175,13 @@ export function getAllSchemas(): Record<
   for (const [_className, registered] of getClasses()) {
     // Skip collection classes - they don't have their own tables
     // Their schemas incorrectly contain collection properties (loaded, options, etc.)
-    if (isCollectionRegistration(_className, registered)) {
+    if (
+      isCollectionRegistration(
+        _className,
+        registered,
+        collectionRegistrationLookup,
+      )
+    ) {
       continue;
     }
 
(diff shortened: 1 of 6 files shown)


packages/core/src/testing/database.ts
```diff
@@ -71,6 +71,153 @@ function isSTIChild(className: string): boolean {
   return stiBaseName !== className;
 }
 
+type RegisteredSchemaClass = NonNullable<
+  ReturnType<typeof ObjectRegistry.getClass>
+>;
+
+function isSmrtCollectionExtendsName(extendsName: string | undefined): boolean {
```
willgriffin: **Duplication (non-blocking).** These four helpers — `isSmrtCollectionExtendsName`, `resolveRelatedRegistration`, `getCollectionAncestorRegistrations`, `isCollectionRegistration` — are duplicated almost verbatim in `packages/core/src/registry/schema-builder.ts` (lines 24-78); the only difference is `ObjectRegistry.getClass` here vs `findClass` there. A third, manifest-data variant lives in `packages/vitest/src/test-db.ts`. Three copies of the same collection-detection logic will drift. Consider extracting one shared helper (e.g. in the registry module, parameterized by the lookup fn) so a future change to the chain-walking rules only has to land once.

willgriffin (commented):
Reviewed the STI collection test-DB schema fix. The root-cause analysis holds up: collection *subclasses* (e.g. `EmailCollection extends MessageCollection extends SmrtCollection`) slipped past the old guards because those only matched a literal `extends === 'SmrtCollection'`. So a collection stub that shares an STI table (`messages`) was treated as a model schema and could create that table before the STI base, omitting `_meta_type` / `_meta_data` — order-dependent and exactly the bug described. Walking the inheritance chain via `getInheritanceChain` and mapping collection subclasses to their item/STI-base schema is the right approach, and catching the over-broad qualified-include widening (covered by the new collision test) is a nice find.

Validation I ran locally on this branch:
- `packages/core/.../test-database-manifest-schema.test.ts` — 5 passed
- `packages/vitest/.../test-db-from-manifest.test.ts` — 19 passed
- core `src/registry` + `test-database` suites — 81 passed / 3 skipped

No correctness issues found. A few non-blocking maintainability notes inline.

packages/vitest/src/test-db.ts
```diff
@@ -745,6 +749,213 @@ function mergeIndexes(
...
+  return key.slice(0, separatorIndex);
+}
+
+function getObjectPackageName(
+  manifest: ManifestFile,
+  key: string,
+  objectDef: ManifestObjectDef,
+): string | undefined {
+  return (
+    objectDef.packageName || manifest.packageName || getPackageNameFromKey(key)
```
willgriffin: **Fallback ordering (non-blocking, latent).** `objectDef.packageName || manifest.packageName || getPackageNameFromKey(key)`: the key-derived package (`@scope/pkg` from `@scope/pkg:Class`) is *more* specific than the manifest-global `packageName`. If a manifest ever aggregates objects from more than one package without per-object `packageName`, every object would be mislabeled with the manifest's package and the package scoping in `findManifestObject` would silently mismatch. Today `loadManifest` reads a single-package file and real objects all carry `packageName`, so this can't trigger now — but `objectDef.packageName || getPackageNameFromKey(key) || manifest.packageName` would be the more robust order.

packages/core/src/registry/schema-builder.ts
```diff
@@ -19,6 +19,63 @@ import { getClasses, getCollectionTableNames } from './shared-state';
...
+
+function getCollectionAncestorRegistrations(
+  className: string,
+  registered: RegisteredSchemaClass,
+): RegisteredSchemaClass[] {
+  const lookupName = registered.qualifiedName || registered.name || className;
+  const chain = ObjectRegistry.getInheritanceChain(lookupName);
+  const registrations: RegisteredSchemaClass[] = [];
+
+  for (const ancestorName of chain) {
```
willgriffin: **Naming nit (non-blocking).** `getInheritanceChain(lookupName)` returns the full chain *including the class itself* as the last element, so this returns `[...ancestors, self]`, not strictly ancestors. Both callers tolerate it (`isCollectionRegistration`'s `.some()` and `resolveCollectionItemClassName`'s reversed scan), but the name reads as ancestors-only and a future caller could assume self is excluded. A rename — or skipping the self entry — would remove the trap. Same applies to the copy in `testing/database.ts`.

packages/core/src/testing/database.ts
```diff
@@ -71,6 +71,153 @@ function isSTIChild(className: string): boolean {
...
+  }
+
+  const ancestors = getCollectionAncestorRegistrations(
+    className,
+    registered,
+  ).reverse();
+
+  for (const ancestor of ancestors) {
+    if (ancestor.extendsTypeArg) {
+      return ancestor.extendsTypeArg;
```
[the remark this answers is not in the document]
willgriffin: Fixed in 4decaf836. Collection item resolution now prefers a runtime static `_itemClass`, then explicit generics, then same-name item inference, before falling back to inherited collection type args. Added core and manifest-helper regressions for the `EmailAttachmentCollection extends AttachmentCollection` shape.

packages/core/src/testing/database.ts
willgriffin: Addressed in 4decaf836. Extracted the duplicated registry collection-chain logic into `packages/core/src/registry/collection-resolution.ts` and wired both `schema-builder` and `getTestDatabase()` through it.

packages/vitest/src/test-db.ts
willgriffin: Addressed in 4decaf836. `getObjectPackageName()` now prefers per-object package, then the package derived from the qualified key, then the manifest-global package fallback.

packages/core/src/registry/schema-builder.ts
willgriffin: Addressed in 4decaf836. The shared helper is named `getCollectionChainRegistrations()` to reflect that the chain includes the current class, avoiding the ancestors-only implication.

PR #1325, opened 2026-06-02
[codex] Emit uuid ids and relationship columns

willgriffin:
## Summary
- Default generated SMRT ids and relationship id columns to UUID, with an `idType` text opt-out for intentional non-UUID primary keys.
- Reconcile same-package foreign keys to the target model id type while keeping cross-package refs as UUID id columns without FK constraints.
- Add a JSON adapter DDL strategy that keeps DuckDB inline constraints but stores UUID columns as TEXT for stable JS round-tripping.
- Regenerate committed package manifests with UUID-aware DDL and normalized source paths.

## Validation
- `pnpm --filter @happyvertical/smrt-core build`
- focused Vitest UUID/schema/scanner suite
- targeted JSON UUID regression suite
- `pnpm --filter @happyvertical/smrt-core test`
- `pnpm run build`
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm test`
- `git diff --check`

### DIFF
--- a/packages/assets/src/manifest/manifest.json
+++ b/packages/assets/src/manifest/manifest.json
@@ -1,6 +1,6 @@
 {
   "version": "1.0.0",
-  "timestamp": 1780375807201,
+  "timestamp": 1780379236418,
   "packageName": "@happyvertical/smrt-assets",
   "packageVersion": "0.25.8",
   "objects": {
@@ -14,8 +14,11 @@
       "fields": {
         "assetId": {
           "type": "foreignKey",
-          "required": false,
-          "related": "Asset"
+          "required": true,
+          "related": "Asset",
+          "_meta": {
+            "required": true
+          }
         },
         "metaType": {
           "type": "text",
@@ -74,6 +77,11 @@
       "collectionExportName": "AssetAssociationCollection",
       "validationRules": [
         {
+          "field": "assetId",
+          "rule": "required",
+          "fieldType": "foreignKey"
+        },
+        {
           "field": "metaType",
           "rule": "required",
           "fieldType": "text"
@@ -91,7 +99,7 @@
       ],
       "schema": {
         "tableName": "asset_associations",
-        "ddl": "CREATE TABLE IF NOT EXISTS \"asset_associations\" (\n  \"id\" uuid PRIMARY KEY,\n  \"slug\" TEXT NOT NULL,\n  \"context\" TEXT NOT NULL DEFAULT '',\n  \"created_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp,\n  \"updated_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp,\n  \"asset_id\" uuid,\n  \"meta_type\" TEXT NOT NULL,\n  \"meta_id\" TEXT NOT NULL,\n  \"role\" TEXT NOT NULL,\n  \"sort_order\" INTEGER DEFAULT 0\n);",
+        "ddl": "CREATE TABLE IF NOT EXISTS \"asset_associations\" (\n  \"id\" UUID PRIMARY KEY NOT NULL,\n  \"slug\" TEXT NOT NULL,\n  \"context\" TEXT NOT NULL DEFAULT '',\n  \"created_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp,\n  \"updated_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp,\n  \"asset_id\" UUID NOT NULL,\n  \"meta_type\" TEXT NOT NULL,\n  \"meta_id\" TEXT NOT NULL,\n  \"role\" TEXT NOT NULL,\n  \"sort_order\" INTEGER DEFAULT 0\n);",
         "columns": {
           "id": {
             "type": "UUID",
@@ -119,7 +127,7 @@
           },
           "asset_id": {
             "type": "UUID",
-            "notNull": false,
+            "notNull": true,
             "unique": false
           },
           "meta_type": {
@@ -162,7 +170,7 @@
             "unique": true
           }
         ],
-        "version": "4f1d3516"
+        "version": "2baed304"
       }
     },
     "@happyvertical/smrt-assets:AssetAssociationCollection": {
@@ -312,7 +320,7 @@
       "collectionExportName": "AssetAssociationCollectionCollection",
       "schema": {
         "tableName": "asset_association_collections",
-        "ddl": "CREATE TABLE IF NOT EXISTS \"asset_association_collections\" (\n  \"id\" uuid PRIMARY KEY,\n  \"slug\" TEXT NOT NULL,\n  \"context\" TEXT NOT NULL DEFAULT '',\n  \"created_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp,\n  \"updated_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp\n);",
+        "ddl": "CREATE TABLE IF NOT EXISTS \"asset_association_collections\" (\n  \"id\" UUID PRIMARY KEY NOT NULL,\n  \"slug\" TEXT NOT NULL,\n  \"context\" TEXT NOT NULL DEFAULT '',\n  \"created_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp,\n  \"updated_at\" TIMESTAMP NOT NULL DEFAULT current_timestamp\n);",
         "columns": {
           "id": {
             "type": "UUID",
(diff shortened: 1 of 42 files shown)


packages/core/src/utils/naming.ts
```diff
@@ -0,0 +1,37 @@
...
+    return [...parts, pluralizeLib(lastWord)].join('_');
+  }
+
+  return pluralizeLib(word);
+}
+
+/**
+ * Converts a class name to a pluralized snake_case table name.
+ */
+export function classnameToTablename(className: string): string {
```
willgriffin, line 33: **[Medium] Duplicated table-naming helpers — drift risk.** This new file re-implements `toSnakeCase`, `pluralize`, and `classnameToTablename`, but those three already exist **verbatim** in `packages/core/src/utils.ts` (lines 16 / 200 / 323), which remains on this branch. Table-name derivation now has two independent sources of truth.

Because table names and FK-target resolution depend on `classnameToTablename`, any future fix applied to one copy but not the other will silently desynchronize computed table names across code paths (manifest generation here vs. the `utils.ts` consumers), producing schema/FK mismatches that are very hard to trace.

Suggest making this the single canonical implementation and having `utils.ts` re-export from it (or vice-versa) rather than duplicating the bodies. If the split exists to avoid an import cycle, a thin re-export still avoids the cycle without duplicating logic.

packages/core/src/registry/schema-builder.ts
```diff
@@ -755,7 +755,9 @@ export function mapFieldTypeToSQL(
     case 'json':
       return 'JSON';
     case 'foreignKey':
-      return 'TEXT';
+      return 'UUID';
+    case 'crossPackageRef':
```
willgriffin, line 764: **[Low / latent] `crossPackageRef` is unconditionally `UUID` and never reconciled to the target's `idType`.** Same-package `foreignKey` columns get reconciled to the target model's id type by `resolveSamePackageForeignKeyColumnTypes` (`scanner/manifest-generator.ts:488` skips everything that isn't `foreignKey`, and it only iterates same-package `manifest.objects`). So a `@crossPackageRef` pointing at a model declared `idType: 'text'` (non-UUID ids) will get a native `uuid` column on Postgres/DuckDB and reject that model's text ids at insert time.

No model uses `idType: 'text'` today, so this is latent — but it's an unguarded assumption that all cross-package targets are UUID. Worth a doc note, or letting `@crossPackageRef` carry an `idType` hint, so the asymmetry doesn't bite once someone adds a text-id model that's referenced across packages.

packages/core/src/scanner/manifest-generator.ts
```diff
@@ -455,6 +455,152 @@ export class ManifestGenerator {
...
+      })),
+      triggers: [],
+      foreignKeys: [],
+      dependencies: [],
+      version: schema.version,
+      packageName: '',
+    };
+
+    schema.ddl = generator.generateSQL(schemaDefinition, 'postgres');
+    schema.version = createHash('sha256')
```
willgriffin: **[Low] `version` hash recomputed with a different formula after reconciliation.** Here `version` is hashed from `JSON.stringify({ tableName, columns })`, but the generators that originally produced these schemas hash different inputs: `generateCTISchemaFromManifest` uses `{ columns, className }` and `generateManifestSchema` (STI) uses `{ columns, baseClassName, descendants }` (`schema/generator.ts`).

So once a table's FK column type is reconciled, its stored `version` no longer matches what a fresh generation would compute. Nothing currently recomputes-and-compares the manifest `version` (it's just carried into the registry), so no functional break today — but it defeats the field's purpose as a stable fingerprint of a fresh generation and would surface as false drift if a consumer ever recomputes it. Recommend threading `className` (and STI descendants) through and reusing the original hash input here. (Note: the regenerated DDL itself is fine — `refreshManifestSchemaDDL`'s column mapping is byte-identical to `convertManifestColumnsToSchemaColumns`.)

packages/core/src/schema/ddl/index.ts
```diff
@@ -26,6 +28,7 @@ import type {
 const strategies: Record<DatabaseEngine, DDLStrategy> = {
   sqlite: new SQLiteStrategy(),
   duckdb: new DuckDBStrategy(),
+  json: new JsonDuckDBStrategy(),
```
willgriffin, line 31: **[Low] `json` is only partially wired as a first-class engine.** `getDDLStrategy` + the strategies map handle `'json'` correctly, but `generateMultiEngineDDL` (below) still emits only `sqlite`/`duckdb`/`postgres`, and `detectEngine(url, 'json')` returns `'duckdb'`.

The known JSON-adapter paths (`SchemaManager`, vitest `setup.ts`) resolve the engine structurally via `exportTable` and set `'json'` directly, so they're correct — and `detectEngine`'s `json→duckdb` is actually right for *query-level* compatibility (`system/compatibility.ts`). The trap is that any future code which picks a DDL strategy via `detectEngine`/explicit `type: 'json'` (instead of the `exportTable` check) would silently get the native-UUID DuckDB strategy and reintroduce the exact round-trip bug this PR fixes. Worth either adding `json` to `generateMultiEngineDDL` or a comment marking why `detectEngine` intentionally collapses it to `duckdb`.

willgriffin (commented):
## Review: Emit UUID ids and relationship columns (R11)

Solid change overall — the abstract `UUID` SQLData type with per-dialect rendering (native `uuid` on PG, native `UUID` on DuckDB, `TEXT` on SQLite/json) is clean, the `idType: 'text'` opt-out is the right escape hatch, and routing the JSON adapter through a dedicated `json` engine to keep UUIDs as `TEXT` is well reasoned. The `differ` already treats `TEXT`≡`UUID`, so existing PG/DuckDB tables won't be force-migrated. I confirmed the regenerated manifests are internally consistent (same-package FKs reconciled to target id type, plain `text()` fields like `parent_id`/`external_id`/`tenant_id` stay `TEXT`, cross-package `asset_id` → `UUID`).

Findings below are mostly consistency / latent-risk items — none are hard blockers, but the first one and the migration note are worth addressing.

### Additional notes (not line-anchored)

**[Low] `--dry-run` SQL preview shows engine-agnostic raw types.** `generateSchema()`'s return value still uses the no-engine `generateSQL` path (`packages/core/src/schema/utils.ts:154`), which emits raw types (`UUID`, `JSON`, `BOOLEAN`). `smrt db:init --dry-run` prints exactly this, so for a SQLite target the preview now shows `"id" UUID` / FK `UUID` columns while the actually-executed DDL (via `SchemaManager`) renders them as `TEXT`. The preview misrepresents the real DDL for non-Postgres targets — consider threading the resolved target engine into that `generateSQL` call.

**[Question] Migration / fresh-deploy compatibility.** All ids and same-package FKs now default to `UUID` → native `uuid` on PG/DuckDB. Because the differ treats `TEXT`≡`UUID`, *existing* tables keep `TEXT`, but *fresh* PG/DuckDB deploys get native `uuid` columns. That means every model's id — including any caller-supplied `options.id` (`packages/core/src/object.ts:2394`) — must be a valid UUID on those dialects. No model opts into `idType: 'text'`, so this assumes no model ever persists a non-UUID id. Worth confirming that holds for models that accept external/deterministic ids (places `lookupOrCreate`, tags, facts, etc.).

packages/core/src/schema/generator.ts
```diff
@@ -1249,7 +1268,7 @@ export class SchemaGenerator {
       packageName: '',
     };
 
-    const ddl = this.generateSQL(schemaDefinition);
+    const ddl = this.generateSQL(schemaDefinition, 'postgres');
```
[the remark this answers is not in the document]
willgriffin: Fixed in 67b5a4797: manifest schema generation now keeps stored DDL dialect-neutral again, and the regenerated manifests no longer contain PostgreSQL-specific JSONB/lowercase uuid DDL.

packages/core/src/schema/generator.ts
```diff
@@ -65,25 +80,30 @@ export class SchemaGenerator {
       case 'json':
         return 'JSON';
       case 'foreignKey':
-        return 'TEXT'; // Foreign keys are typically TEXT
+        return 'UUID'; // Foreign keys default to UUID, then same-package refs can be reconciled to target idType
```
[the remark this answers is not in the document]
willgriffin: Fixed in 67b5a4797: runtime schema generation now passes ObjectRegistry context into SchemaGenerator and reconciles foreignKey columns against the referenced target idType. Added runtime/text-id coverage in uuid-schema-generator.test.ts.

packages/core/src/utils/naming.ts
willgriffin: Fixed in 67b5a4797: packages/core/src/utils.ts now imports/re-exports the canonical helpers from utils/naming.js, removing the duplicate implementations while keeping existing public imports working.

packages/core/src/registry/schema-builder.ts
willgriffin: Fixed in 67b5a4797: crossPackageRef now supports an idType hint, the scanner preserves it in metadata, and schema generation/schema-builder render text-id cross-package refs as TEXT.

packages/core/src/scanner/manifest-generator.ts
willgriffin: Fixed in 67b5a4797: refreshManifestSchemaDDL now recomputes manifest versions with the same CTI/STI hash inputs used by fresh schema generation, including STI base and descendants.

packages/core/src/schema/ddl/index.ts
willgriffin: Fixed in 67b5a4797: generateMultiEngineDDL now includes json, and detectEngine has comments documenting why type/json URLs intentionally collapse to duckdb for query compatibility while DDL callers should request json explicitly or detect exportTable.

packages/core/src/testing/database.ts
```diff
@@ -244,7 +246,7 @@ export async function getTestDatabase(
           );
 
     // Generate DDL using generateSQL() - the single source of truth
-    const ddl = schemaGenerator.generateSQL(schema);
+    const ddl = schemaGenerator.generateSQL(schema, 'sqlite');
 
     try {
       await db.query(ddl);
```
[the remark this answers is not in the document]
willgriffin: Fixed in 67b5a4797: getTestDatabase now selects the json DDL strategy for JSON-like adapters via type/exportTable, including index generation. Added JSON adapter DDL coverage in test-database-manifest-schema.test.ts.

packages/assets/src/manifest/manifest.json
```diff
@@ -13,11 +13,9 @@
...
         "assetId": {
-          "type": "text",
-          "required": true,
-          "_meta": {
-            "required": true
-          }
+          "type": "foreignKey",
+          "required": false,
+          "related": "Asset"
         },
```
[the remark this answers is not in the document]
willgriffin: Fixed in 67b5a4797: @foreignKey options now override initializer-based requiredness in the scanner, and the regenerated AssetAssociation manifest preserves assetId required: true plus validation metadata.

packages/assets/src/manifest/manifest.json
```diff
@@ -125,8 +118,8 @@
             "default": "current_timestamp"
           },
           "asset_id": {
-            "type": "TEXT",
-            "notNull": true,
+            "type": "UUID",
+            "notNull": false,
             "unique": false
           },
```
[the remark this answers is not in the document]
willgriffin: Fixed in 67b5a4797: the regenerated asset_associations schema now has asset_id UUID NOT NULL, matching @foreignKey("Asset", { required: true }).

PR #1341, opened 2026-06-03
[codex] Build SMRT dev knowledge foundation

willgriffin:
## Summary

Builds `@happyvertical/smrt-dev-mcp` into a deterministic SMRT + HappyVertical SDK knowledge layer and migrates package expert docs to canonical `AGENTS.md` with one-line `CLAUDE.md` shims.

Adds shared knowledge indexing/freshness/context builders, CLI and MCP surfaces for review and architecture context, deterministic review/architecture outputs, markdown/json checker output, local Lefthook knowledge freshness checks, stdio MCP smoke coverage, and docs for local/manual model-assisted prompt-bundle workflows.

Also ships a harness-agnostic downstream `smrt-review` agent skill in the package payload and exposes it through `list-agent-skills` / `get-agent-skill`, so Codex, Claude, or another MCP harness can fetch the exact review procedure before calling `smrt-review` on a project diff.

## Validation

- `pnpm --filter @happyvertical/smrt-dev-mcp test` (63 passed)
- `pnpm --filter @happyvertical/smrt-dev-mcp build`
- `pnpm --filter @happyvertical/smrt-dev-mcp verify:pack`
- `npm pack --dry-run --json` from `packages/smrt-dev-mcp` confirms `agent-skills/` is included
- `pnpm --filter @happyvertical/smrt-cli test` (347 passed, 4 skipped)
- `pnpm --filter @happyvertical/smrt-cli build`
- `pnpm run check:standards`
- `pnpm knowledge:check --strict --format markdown`
- branch pre-push hook: `format-check`, `knowledge-check`, `validate-all-workflows`

## Notes

Model-assisted review remains local/manual. CI and hooks only use deterministic checks; prompt bundles and bundled skills are the integration boundary for Codex, Claude, or another local model workflow.


packages/smrt-dev-mcp/vite.config.ts
```diff
@@ -10,9 +10,12 @@ export default defineConfig({
...
   build: {
     lib: {
-      entry: 'src/index.ts',
+      entry: {
+        index: 'src/index.ts',
+        knowledge: 'src/knowledge/index.ts',
+      },
       formats: ['es'],
-      fileName: () => 'index.js',
+      fileName: (_format, entryName) => `${entryName}.js`,
```
willgriffin, line 27: **🔴 Blocker — the published `bin` will lose its shebang, breaking the only documented launch path.**

`src/index.ts:1` starts with `#!/usr/bin/env node` and `package.json` maps `bin.smrt-dev-mcp → ./dist/index.js`. Rollup/Vite **lib mode strips a leading `#!`** (it parses as a comment), and there's no `rollupOptions.output.banner` here. The README's only integration snippet is `{"command":"npx","args":["-y","@happyvertical/smrt-dev-mcp"]}`, which executes `dist/index.js` directly via its shebang — so the server won't start on Claude, Codex, or any other harness.

The sibling bin package already solves this — `packages/app-cli/vite.config.ts`:
```ts
output: {
  entryFileNames: '[name].js',
  // Preserve the shebang on the bin entry — Rollup strips it by default
  banner: (chunk) => (chunk.fileName === 'index.js' ? '#!/usr/bin/env node' : ''),
}
```
**Not caught by tests** — `src/mcp-stdio.test.ts` launches via `tsx src/index.ts` (source), never the built `dist/index.js`. Fix: add the banner, and add a smoke test that spawns the built bin.

packages/smrt-dev-mcp/vite.config.ts
```diff
@@ -10,9 +10,12 @@ export default defineConfig({
   ],
   build: {
     lib: {
-      entry: 'src/index.ts',
+      entry: {
+        index: 'src/index.ts',
+        knowledge: 'src/knowledge/index.ts',
```
willgriffin: **🟠 Major — `rollupTypes:true` with two lib entries risks not emitting `dist/knowledge.d.ts`.**

The CLI consumes the subpath: `packages/cli/src/commands/dev-knowledge.ts` imports from `@happyvertical/smrt-dev-mcp/knowledge`, and `package.json` maps `./knowledge → dist/knowledge.d.ts`. vite-plugin-dts `rollupTypes` is oriented around a single rolled entry; the `secrets` package hit this exact hazard and had to switch to `rollupTypes:false` + `entryRoot` (documented in `packages/secrets/vite.config.ts:6-9`).

`prepack`'s `verify:pack` asserts every `exports` `types` path exists in the tarball, so a broken state fails loudly rather than shipping — but the package may be **unable to publish** until this is fixed. Please run `pnpm --filter @happyvertical/smrt-dev-mcp build && npm run verify:pack` and confirm `dist/knowledge.d.ts` is produced; if not, mirror the `secrets` config.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -88,7 +100,129 @@ const TOOLS = [
...
+      properties: {
+        rootDir: { type: 'string' },
+        changedFiles: { type: 'array', items: { type: 'string' } },
+        focus: { type: 'string' },
+        documentation: { type: 'string' },
+      },
+    },
+  },
+  {
+    name: 'smrt-review',
```
willgriffin, line 156: **🟠 Major — non-Claude harnesses have no path to *activate* the skill; plus a tool/skill name collision.**

The skill is only the return payload of `list-agent-skills`/`get-agent-skill`. Nothing signals to a generic MCP client / Cursor / Codex that it should call them *before* reviewing — the "fetch the skill first" instruction lives only in README/SKILL.md the model may never read. In Claude the skill model triggers on its description; elsewhere there's no trigger.

Fix: fold "first call `get-agent-skill { name: 'smrt-review' }`" into **this tool's `description`** (every harness reads tool descriptions).

Separately, this MCP tool `smrt-review` shares a name with the bundled **skill** `smrt-review` (`agent-skills.ts:26`), and the skill body instructs the model to "call MCP tool `smrt-review`". For a harness that flattens skills+tools into one list this is genuinely confusing — consider renaming the skill to e.g. `smrt-code-review`.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -88,7 +100,129 @@ const TOOLS = [
...
+      properties: {
+        rootDir: { type: 'string' },
+        idea: { type: 'string' },
+        documentation: { type: 'string' },
+        focus: { type: 'string' },
+      },
+    },
+  },
+  {
+    name: 'list-agent-skills',
```
willgriffin, line 203: **🟠 Major (design) — skills/prompt-bundles are documents, but the server only exposes `tools`.**

`capabilities: { tools: {} }` + stdio is the correct lowest-common-denominator for *execution*, and JSON-text content blocks are universally consumable — so this isn't a blocker. But a "knowledge/review layer" whose primary artifacts are procedures and context is exactly what MCP `prompts` and `resources` exist for. Exposing the bundled skill via `ListPrompts`/`GetPrompt` (in addition to these tools) would give skill-aware-but-non-Claude harnesses a native way to surface it to users. Worth considering for the "any harness" goal; not required for merge.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -140,6 +274,64 @@ async function main() {
           result = await introspectProject(args as any);
           break;
 
+        case 'reflect-knowledge': {
+          const index = await buildKnowledgeIndex(args as any);
```
willgriffin, line 352: **⚪ Minor (perf) — `reflect-knowledge` walks the whole monorepo twice.**

`buildKnowledgeIndex(args)` here, then `checkKnowledgeFreshness(args)` on the next line — but `checkKnowledgeFreshness` calls `buildKnowledgeIndex` internally. So one call discovers all packages, reads every `package.json`/manifest, walks every `src/` tree, and scans `node_modules/@happyvertical` **twice**. Add a `checkKnowledgeFreshnessFromIndex(index, opts)` variant and pass the already-built index.

willgriffin (commented):
## Multi-angle review — "usable from any harness"

Reviewed from three angles (harness-agnostic design, correctness/bugs, packaging + the AGENTS.md/CLAUDE.md migration) against the post-merge head `9339d99`. Inline comments below carry the detail; this is the index.

**Verdict:** the migration half (AGENTS.md canonical + `@AGENTS.md` shim across all 48 packages + root) is clean and consistent — verified package-by-package, all pass. The deterministic knowledge engine is solid. But the headline goal is **not yet met**: the one documented launch path (`npx -y @happyvertical/smrt-dev-mcp`) is broken at build time, and the bundled "skill" only auto-activates under Claude.

### 🔴 Blocker
- **Stripped shebang → documented `npx` launch fails on every harness** — `vite.config.ts` (see inline).

### 🟠 Major
- **`rollupTypes:true` + 2 lib entries may not emit `dist/knowledge.d.ts`** (the CLI imports `/knowledge`) — `vite.config.ts`.
- **Skill activation is Claude-shaped** — only delivered as tool payloads; no trigger / no MCP `prompts`/`resources` for non-Claude harnesses — `src/index.ts`.
- **`stale-docs-codex-command` guard is inert** (case-sensitive `Codex`) — `knowledge/index.ts:224`.

### 🟡 Correctness (medium)
- `getChangedFiles` default `HEAD...HEAD` is always empty — `knowledge/index.ts:1391`.
- Stale-pattern scan asymmetry (`--changed` has no file-type whitelist) — `knowledge/index.ts:897`.
- `smrt-review` `mode` param accepted but ignored — `knowledge/index.ts:426`.

### ⚪ Minor / cleanup
- Inline on: absolute-vs-relative issue paths, `selectPackages` substring matching, duplicated invariant enforcement, `dedupePackages` order-fragility, `agent-skills.ts` path math, unused `@happyvertical/logger` dep, README "Codex-compatible frontmatter", `@AGENTS.md` shim assumptions.
- **No inline anchor (lands in a diff gap):** `SERVER_VERSION = '0.1.0'` (`src/index.ts:29`) is reported to MCP clients in the handshake but `package.json` is `0.27.0` — inject the real version at build time instead of hardcoding.

### ✅ Confirmed good
`npm pack --dry-run` ships `agent-skills/smrt-review/SKILL.md` + references; all 48 packages + root migrated to the exact `@AGENTS.md` shim; no stale rich CLAUDE.md anywhere; `docs:claude`→`docs:agents` migrated with a deprecated alias + test; no `.npmignore` blocks the skill; stdio + tools + JSON-text is the right neutral execution baseline.

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+    message: 'Stale SMRT core namespace reference found',
+  },
+  {
+    code: 'stale-field-helper-import',
+    pattern: /@happyvertical\/smrt-core\/fields/,
+    message: 'Stale field-helper import path found',
+  },
+  {
+    code: 'stale-docs-codex-command',
+    pattern: new RegExp(`docs:${'Codex'}|docs-${'Codex'}|\\.${'Codex'}`),
```
willgriffin: **🟠 Major — this anti-Codex-coupling guard can never fire.**

The compiled pattern is `/docs:Codex|docs-Codex|\.Codex/` — **case-sensitive, capital `C`** (no `i` flag). Real downstream references are lowercase: `docs:codex`, `docs-codex`, `.codex/`, `codex-command`. So the guard meant to keep harness-specific commands out of the docs (on-mission for "any harness") is inert. (The `${'Codex'}` split is just self-match avoidance, but the casing defeats the purpose.) Fix:
```ts
pattern: new RegExp('docs:codex|docs-codex|\\.codex|codex-command', 'i'),
```
and add a unit test feeding a `docs:codex` fixture and asserting the issue fires (current tests only check the tool list contains the skill tools, not this rule's matching).

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+  }
+  lines.push('');
+  return lines.join('\n');
+}
+
+function getChangedFiles(rootDir: string, base = 'HEAD'): string[] {
+  try {
+    const output = execFileSync(
+      'git',
+      ['diff', '--name-only', `${base}...HEAD`],
```
willgriffin: **🟡 Medium — default base `HEAD` makes this a no-op on a clean worktree.**

`base='HEAD'` → `git diff --name-only HEAD...HEAD`, and the three-dot diff of a ref with itself is always empty. On a clean checkout (CI, freshly checked-out branch) it then falls through to the working-tree+staged fallback, which is also empty. So `buildReviewContext`/`smrt-review` invoked with no `changedFiles` silently see an empty diff and fall back to default package selection — the "review context for this diff" is empty with no warning. `diffKnowledgeIndex` (which passes a real base) is fine; only the default path is wrong. Fix: when no base is given, go straight to the working-tree+staged diff rather than `HEAD...HEAD`.

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+    features.add('uuidColumns');
+  }
+  return [...features].sort();
+}
+
+function findStalePatternIssues(
+  rootDir: string,
+  changedFiles?: string[],
+): KnowledgeIssue[] {
+  const candidates =
```
willgriffin, line 919: **🟡 Medium — `--changed` scans file types the full-scan path never does (asymmetric false positives).**

When `changedFiles` is provided, every file is scanned directly with **no file-type filter**; the full-scan branch (below) restricts candidates to `AGENTS.md`/`README.md`/`docs/content/*.md`. So stale patterns like `@have/` or `@smrt/core` in a `.ts` comment or a `package.json` version range are flagged only in `--changed` mode — exactly the mode the pre-commit hook uses. This is being patched reactively one type at a time (`pnpm-lock.yaml` at line 925, and the `CHANGELOG.md` skip just added at line 926 after the main-merge tripped it). Replace the deny-list with the same positive whitelist the full-scan path uses, so both modes agree.

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+
+export async function smrtReview(
+  options: ContextSelectorOptions & {
+    mode?: 'findings' | 'prompt-bundle' | 'both';
+  } = {},
+): Promise<SmrtReviewResult> {
+  const context = await buildReviewContext(options);
+  return {
+    ...context,
+    mode: options.mode ?? 'both',
```
willgriffin: **🟡 Medium — `mode` is accepted but never honored (contract mismatch).**

The input schema advertises `mode: 'findings' | 'prompt-bundle' | 'both'`, but this just spreads the full context and tacks on `mode`. `mode:'findings'` still returns `promptBundle`; `mode:'prompt-bundle'` still returns `deterministicFindings`. Every harness gets the full payload regardless of what it requested. Either filter the output by `mode`, or drop the param from the schema (`index.ts:160-166`).

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+        });
+      }
+    }
+
+    if (!pkg.files.includes('AGENTS.md')) {
+      issues.push({
+        severity: 'error',
+        code: 'package-files-missing-agents',
+        message: 'package.json files allowlist must include AGENTS.md',
+        file: packageJsonPath,
```
willgriffin: **⚪ Minor — these two issue codes emit absolute `file` paths; every other code emits root-relative.**

`package-files-missing-agents` (here) and `package-files-missing-claude-shim` (line 333) set `file: packageJsonPath` (absolute), while `missing-agents-md`, `missing-claude-shim`, `claude-not-shim`, and `package-files-entry-missing` all use `relative(index.rootDir, ...)`. Any consumer that link-resolves `issue.file` against the root gets wrong output for these two. Wrap both in `relative(index.rootDir, packageJsonPath)`.

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+): KnowledgePackage[] {
+  const selected = new Set<KnowledgePackage>();
+  for (const pkg of selectPackagesForFiles(index, options.changedFiles ?? [])) {
+    selected.add(pkg);
+  }
+
+  const text = (options.text ?? '').toLowerCase();
+  if (text) {
+    for (const pkg of index.smrtPackages) {
+      const packageKey = pkg.name.replace('@happyvertical/smrt-', '');
```
willgriffin, line 1265: **⚪ Minor — substring matching on short package keys over-selects packages.**

`packageKey` becomes `core`, `ads`, `chat`, `jobs`, `cli`, `tags`, `voice`, etc. `text.includes(packageKey)` then matches unrelated prose — "the user **chat**s with an agent" pulls in `smrt-chat`; "schedule **jobs**" pulls in `smrt-jobs`. Same for `selectSdkPackages` short names (`ai`, `sql`, `json` → "the J**SON** payload" matches `@happyvertical/json`). The result is a noisier prompt bundle. Use word-boundary matching: `new RegExp(\`\\b${packageKey}\\b\`, 'i').test(text)`.

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+export async function checkKnowledgeFreshness(
+  options: CheckKnowledgeFreshnessOptions = {},
+): Promise<KnowledgeFreshnessResult> {
+  const index = await buildKnowledgeIndex(options);
+  const issues: KnowledgeIssue[] = [];
+  const changedFiles = options.changed
+    ? getChangedFiles(index.rootDir)
+    : undefined;
+
+  for (const pkg of index.packages.filter((item) => item.kind !== 'sdk')) {
```
willgriffin, line 283: **⚪ Minor (maintainability) — the shim invariant is enforced in two places that have already diverged.**

This block and `scripts/check-standards.mjs` both require AGENTS.md present, `CLAUDE.md === '@AGENTS.md'`, and `files` includes both. But they differ on scope: `check-standards.mjs` enumerates *every* `packages/*` dir, while this iterates only discovered `kind !== 'sdk'` packages; and `check-standards.mjs` also requires `files` to include `dist` (with exemptions) whereas this has no `dist` check. Two sources of truth in two languages → a future rule change must touch both. Factor into one shared module, or at minimum cross-reference each in a comment.

packages/smrt-dev-mcp/src/knowledge/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+    );
+  } catch {
+    return [];
+  }
+}
+
+function dedupePackages(packages: KnowledgePackage[]): KnowledgePackage[] {
+  const byName = new Map<string, KnowledgePackage>();
+  for (const pkg of packages) {
+    if (!byName.has(pkg.name) || pkg.kind !== 'sdk') {
```
willgriffin: **⚪ Minor (robustness) — `dedupePackages` collision rule depends on insertion order.**

`if (!byName.has(pkg.name) || pkg.kind !== 'sdk')` keeps the workspace entry today only because workspace packages are pushed before SDK packages. If the input order ever flips, an SDK entry would overwrite a workspace entry. An order-independent form expresses the intent directly:
```ts
if (!byName.has(pkg.name) || byName.get(pkg.name).kind === 'sdk') byName.set(pkg.name, pkg);
```

packages/smrt-dev-mcp/src/agent-skills.ts
```diff
@@ -0,0 +1,67 @@
...
+        ? []
+        : skill.references.map((path) => ({
+            path,
+            content: readPackageFile(path),
+          })),
+  };
+}
+
+function readPackageFile(relativePath: string): string {
+  const packageRoot = dirname(dirname(fileURLToPath(import.meta.url)));
```
willgriffin: **⚪ Minor — `dirname(dirname(...))` is correct but silently coupled to the build output depth, with no comment or guard.**

Today `agent-skills.ts` is bundled into `dist/index.js`, so `import.meta.url` → `<pkg>/dist/index.js` → `dirname` ×2 → `<pkg>`, and `agent-skills/` is a sibling of `dist/`. Correct. But if the build ever emits one level deeper (e.g. `dist/esm/index.js`, as `app-cli` does with `dist/bin/...`), this resolves to `dist/` and every `readFileSync` throws `ENOENT` with no context — and no test would catch it (no test loads the skill through the built layout). Add a one-line comment documenting the depth assumption, and ideally a test that calls `getAgentSkill({name:'smrt-review'})` against the built `dist/`.

packages/smrt-dev-mcp/package.json
```diff
@@ -12,13 +12,18 @@
...
     }
   },
   "files": [
     "CLAUDE.md",
-    "LICENSE",
     "README.md",
-    "dist"
+    "dist",
+    "AGENTS.md",
+    "agent-skills"
```
willgriffin, line 26: **⚪ Minor (packaging housekeeping)** — `agent-skills` in the `files` allowlist is correct and `npm pack --dry-run` confirms `agent-skills/smrt-review/SKILL.md` + `references/` ship. ✅

While here: `@happyvertical/logger` in `dependencies` is **not imported anywhere in `src/`** (it appears only as a string inside the `SDK_PACKAGE_NAMES` data set in `knowledge/index.ts`). `@happyvertical/ai`, `@happyvertical/files`, and `@happyvertical/utils` also have no direct `import` in `src/` — they're externalized in the vite build but unused directly. Audit and drop the ones not needed (or demote to peer), starting with `logger`.

packages/smrt-dev-mcp/README.md
```diff
@@ -26,9 +28,46 @@ Add to your `.mcp.json` or Claude Desktop config:
...
+```json
+{
+  "name": "get-agent-skill",
+  "arguments": {
+    "name": "smrt-review"
+  }
+}
+```
+
+The returned `skillMarkdown` is plain Markdown with Codex-compatible
```
willgriffin: **⚪ Minor — "Codex-compatible frontmatter" is inaccurate and name-drops one harness in a doc about being harness-agnostic.**

The frontmatter (`SKILL.md`: `name` + `description`) is the **Anthropic Agent Skills** `SKILL.md` schema, not a Codex convention (Codex/`AGENTS.md` has no SKILL.md frontmatter concept). Call it "plain Markdown with YAML frontmatter (`name`/`description`); the body is harness-neutral prose," and note that non-skill-aware harnesses should ignore the frontmatter and read the body. Same wording appears in `packages/smrt-dev-mcp/AGENTS.md:24`.

CLAUDE.md
```diff
@@ -1,113 +1 @@
...
-- **ObjectRegistry on globalThis**: singleton via `globalThis.__smrtRegistry*` — survives HMR
-- **Lazy table creation**: tables created on first DB op, not on `initialize()` — safe for SSR
-- **Svelte typechecking**: packages with `/svelte` subpaths must run `svelte-check` in `typecheck` after `tsc`; `tsc` alone treats `.svelte` imports as ambient `Component<any>`.
-- **TypeScript & Svelte 5 Performance**: Avoid using inline intersected generic types (like `Asset & { id: string }`) in component `$props()`. It causes infinite-loop-like recursion during type evaluation. Export an explicit interface (e.g., `interface PersistedAsset extends Asset { id: string }`) instead.
-- **Editing `packages/core/src/scanner/*` or `src/schema/generator.ts` requires a rebuild**: the vite-plugin loads these modules from `dist/` deterministically (sniffing `.ts`/`.js` via `import.meta.url` was non-deterministic under tsx and repeatedly broke publishes — see #1139). Run `pnpm build` in core, or keep `pnpm dev` / `pnpm build:watch` running, otherwise consumers won't see your scanner/generator edits reflected in their manifests.
+@AGENTS.md
```
willgriffin, line 1: **🟡 Watch — the `@AGENTS.md` shim depends on a Claude-Code-specific include, with no fallback or test.**

This rehydrates the full doc **only under harnesses that resolve `@`-imports** (Claude Code does; Codex reads `AGENTS.md` natively so the shim is moot for it). But any tool that reads `CLAUDE.md` as plain markdown without processing `@`-imports now gets the literal two-word string where it previously got the full doc — a strict content regression for such consumers, relative to the old fat `CLAUDE.md`. The freshness checker enforces `claudeContent.trim() === '@AGENTS.md'` but nothing tests that the import actually *resolves*. Given the "any harness" framing: add a doc note (and ideally a CI assertion) pinning the "Claude Code resolves bare `@AGENTS.md`" assumption, and state in `AGENTS.md` that non-`@`-import harnesses should read `AGENTS.md` directly. (Blast radius is small per the known consumer set, but worth acknowledging.)

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -195,7 +387,14 @@ async function main() {
   });
 }
 
-main().catch((error) => {
-  console.error(`[${SERVER_NAME}] Fatal error:`, error);
-  process.exit(1);
-});
+function isEntrypoint(): boolean {
+  const entry = process.argv[1];
+  return Boolean(entry && import.meta.url === pathToFileURL(entry).href);
```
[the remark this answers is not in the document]
willgriffin: Addressed in aa4dd4e37. `isEntrypoint()` now compares real paths for `import.meta.url` and `process.argv[1]`, and `verify:pack` starts the built MCP server through a symlinked bin path to cover the installed npm/pnpm launch path.

packages/smrt-dev-mcp/src/tools/introspect-project.ts
```diff
@@ -114,11 +114,11 @@ export async function introspectProject(
         // Pattern 1: TypeScript property declarations (new decorator pattern)
         // Example: name: string = ''; or price: number = 0.0;
         const tsFieldMatches = content.matchAll(
-          /^\s*(?:@field\([^)]*\)\s*)?(\w+):\s*(\w+)(?:<[^>]+>)?\s*=/gm,
+          /^\s*(?:@\w+\([^)]*\)\s*)*(\w+)(?::\s*([\w| ]+)(?:<[^>]+>)?)?\s*=\s*([^;\n]+)/gm,
         );
```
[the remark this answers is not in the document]
willgriffin: Addressed in aa4dd4e37. The introspection regex now captures TypeScript modifiers separately, skips non-persisted static/private/protected/declare/abstract members, and has a regression test for modifier handling.

docs/content/gnode.md
```diff
@@ -55,9 +55,9 @@ See the [town template](../create-gnode/templates/town/) for complete implementa
 
 ## Related Packages
 
-- [@have/smrt](../smrt) - Core SMRT framework
-- [@have/sql](../sql) - Database abstraction layer
-- [@have/create-gnode](../create-gnode) - CLI generator for creating gnodes
+- [@happyvertical/smrt-core](../smrt) - Core SMRT framework
+- [@happyvertical/sql](../sql) - Database abstraction layer
+- [@happyvertical/smrt-gnode](../create-gnode) - CLI generator for creating gnodes
```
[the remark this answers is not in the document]
willgriffin: Addressed in aa4dd4e37. The related package bullet now labels and links `create-gnode` as the generator instead of presenting it as `@happyvertical/smrt-gnode`.

packages/smrt-dev-mcp/vite.config.ts
willgriffin: Addressed in aa4dd4e37. The Vite build adds a shebang banner for the built bin, the source shebang was removed to avoid duplication, and `verify:pack` checks the built entry plus launches it through a symlinked bin path.

packages/smrt-dev-mcp/vite.config.ts
willgriffin: Addressed in aa4dd4e37. The d.ts config now uses `rollupTypes:false` with `entryRoot`, adds a public `src/knowledge.ts` entry, and `verify:pack` confirms `dist/knowledge.d.ts` exists.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Addressed in aa4dd4e37. The bundled skill is now `smrt-code-review` to avoid the `smrt-review` tool collision, and the `smrt-review` tool description explicitly tells downstream harnesses to load the skill first via `get-agent-skill` or MCP prompt/resource.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Addressed in aa4dd4e37. The server now advertises MCP `prompts` and `resources` capabilities and exposes the `smrt-code-review` skill through both native prompt and resource APIs, with stdio smoke coverage.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Addressed in aa4dd4e37. Added `checkKnowledgeFreshnessFromIndex()` and changed `reflect-knowledge` to reuse the already-built index instead of walking the repo twice.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. The stale Codex-doc pattern is case-insensitive, includes `codex-command`, avoids source self-matches through the changed-file whitelist, and has a lowercase `docs:codex` regression test.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. `getChangedFiles()` now only performs the three-dot diff when a base is explicitly provided; default callers use working-tree plus staged changes, while `diffKnowledgeIndex()` still passes an explicit base.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. Changed-mode stale scanning now uses the same positive markdown/doc allowlist as full scanning, with a regression test proving changed TypeScript files are ignored for stale doc patterns.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. `smrtReview()` now honors `mode` by returning findings only, prompt bundle only, or both, and the behavior has test coverage.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. Both package-files allowlist issues now emit root-relative package.json paths, with a regression test.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. Package and SDK short-name matching now uses token-boundary matching instead of raw substring matching.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. Added a cross-reference comment in the knowledge checker pointing to `scripts/check-standards.mjs`, and the standards check passes with both paths enforcing the same AGENTS/CLAUDE invariants.

packages/smrt-dev-mcp/src/knowledge/index.ts
willgriffin: Addressed in aa4dd4e37. `dedupePackages()` now deterministically prefers an existing non-SDK package over SDK discoveries regardless of insertion order.

packages/smrt-dev-mcp/src/agent-skills.ts
willgriffin: Addressed in aa4dd4e37. Added a package-root guard/comment around the runtime layout assumption, and the new build verifier launches the built server through a symlink and calls `get-agent-skill` to prove the packaged skill layout works.

packages/smrt-dev-mcp/package.json
willgriffin: Addressed in aa4dd4e37. Removed the unused direct runtime dependencies on `@happyvertical/ai`, `@happyvertical/files`, `@happyvertical/logger`, `@happyvertical/sql`, and `@happyvertical/utils`; the lockfile importer was updated.

packages/smrt-dev-mcp/README.md
willgriffin: Addressed in aa4dd4e37. README and package AGENTS docs now describe the skill as plain Markdown with YAML frontmatter and harness-neutral body text, and references were updated to `smrt-code-review`.

CLAUDE.md
willgriffin: Addressed in aa4dd4e37. Root AGENTS.md now states that harnesses which do not resolve Claude Code `@AGENTS.md` shims should read `AGENTS.md` directly; the standards/knowledge checks continue to enforce the one-line shim.

PR #1343, opened 2026-06-03
Add downstream SMRT domain knowledge artifacts

willgriffin:
## Summary
- add shared domain knowledge contract types and core smrt-knowledge artifact generation
- teach smrt-dev-mcp to prefer domain artifacts, check freshness, and expose domain CLI/MCP contexts, resources, and prompts
- add guarded opt-in SvelteKit knowledge API route generation plus template/docs/standards updates

## Validation
- pnpm --dir packages/core exec vitest run src/knowledge.test.ts src/vite-plugin/sveltekit-generator.test.ts
- pnpm --dir packages/smrt-dev-mcp exec vitest run src/knowledge/index.test.ts src/index.test.ts src/mcp-stdio.test.ts
- pnpm --dir packages/cli exec vitest run src/commands/__tests__/dev-knowledge.test.ts
- pnpm --dir packages/template-sveltekit exec vitest run __tests__/copyTemplate.test.ts
- pnpm --dir packages/config exec tsc -p tsconfig.json --noEmit
- pnpm --dir packages/core typecheck
- pnpm --dir packages/smrt-dev-mcp exec tsc -p tsconfig.json --noEmit
- pnpm --dir packages/cli exec tsc -p tsconfig.json --noEmit
- pnpm --dir packages/core build && pnpm --dir packages/core verify:exports
- pnpm --dir packages/smrt-dev-mcp build
- pnpm --dir packages/cli build
- pnpm check:standards
- pnpm knowledge:check --strict --format markdown


packages/core/src/vite-plugin/sveltekit-generator.ts
```diff
@@ -1640,6 +1696,98 @@ ${optionsLoad}  const result = ${buildActionInvocationExpression(
...
+const ARTIFACT_PATHS = [
+  '.smrt/smrt-knowledge.json',
+  'dist/smrt-knowledge.json',
+];
+const INCLUDE_DOCS_BY_DEFAULT = ${JSON.stringify(includeDocs)};
+const INCLUDE_PROMPTS_BY_DEFAULT = ${JSON.stringify(includePrompts)};
+const REQUIRE_ADMIN = ${JSON.stringify(requireAdmin)};
+
+export const GET: RequestHandler = async ({ locals, url }) => {
+  if (!dev && REQUIRE_ADMIN && !isKnowledgeAdmin(locals as Record<string, any>)) {
```
willgriffin, line 1750: **🟠 Major (security) — when `requireAdmin: false`, this serves a full reconnaissance payload to anonymous users in production.**

Past this guard (and `requireAdmin:false` removes it entirely in prod) the response is `sanitizeKnowledgeArtifact(...)`, which only deletes `agentDoc` and empties `prompts`. Everything else is returned unauthenticated: the complete object/data model (every object, `tableName`, fields, relationships, methods), the full API/CLI/MCP/AI **surface map with paths + methods** (an attacker's endpoint inventory), and **exact dependency versions** (`dependencies`) → direct CVE targeting. There's no rate limiting and no caching guard.

The default is safe (`requireAdmin !== false` → true), and the risk is documented — but the public escape hatch should be safe-by-construction, not operator-dependent. Suggest: when `requireAdmin` is false, also strip `dependencies`/`sourceHashes` from the anonymous response, and emit a generated startup `console.warn` that the route is PUBLIC. The auth guard itself (`isKnowledgeAdmin`/`rolesFrom`) is correctly fail-closed — good.

willgriffin (commented):
## Multi-angle review — domain knowledge artifacts (stacked on the merged #1341)

This PR is one substantive commit (`455e390`) on top of main: a core **domain-knowledge layer** (contract types, artifact builder, a committed `smrt-knowledge.json`), a **guarded opt-in SvelteKit knowledge API route generator**, and `smrt-dev-mcp` now exposing **MCP `resources` + `prompts`** (which directly implements the #1341 feedback — nice). Reviewed from four angles: security, correctness, MCP/harness-agnostic, and packaging/config/templates.

**Verdict:** solid and well-documented work. The security default posture is genuinely "guarded opt-in" and the auth guard is correctly fail-closed. But there are a few things to fix before merge — the headline ones below.

### 🟠 Major (fix before merge)
- **The committed `smrt-knowledge.json` is already stale** (`packageVersion: 0.27.0` vs `package.json` `0.27.1` after the main-merge), and `checkDomainKnowledgeArtifact` raw-hashes `package.json` → emits `stale-domain-knowledge` under `--strict`. So this PR's own `pnpm knowledge:check --strict` gate now fails. Worse, every changeset version bump re-stales it → recurring churn. Recommend gitignoring it like its sibling `static-manifest.json`. — `core/src/manifest/smrt-knowledge.json`
- **The new MCP prompts can't receive their key argument from a compliant client.** `domain-code-review`/`domain-architecture` declare no `arguments` and pass `request.params.arguments` (spec type = `Record<string,string>`) into `buildReviewContext` expecting `changedFiles: string[]` → throws `changedFiles.map is not a function` when a client sends it the only way the spec allows. — `smrt-dev-mcp/src/index.ts:385`
- **Public (`requireAdmin:false`) knowledge route serves a recon payload** (full data model + endpoint map + exact dependency versions) unauthenticated, uncached, no rate limit; the sanitizer only strips docs/prompts. — `sveltekit-generator.ts`
- **The route reads its artifact from `process.cwd()` at request time** — unreliable on Vercel/Cloudflare/static adapters (404 in prod), blocking sync FS, no caching. — `sveltekit-generator.ts`

### 🟡 / ⚪ Minor & nits (inline)
Determinism churn (`generatedAt`, vite path lacks the preserve-guard), manifest-hash instability, `include:[]` → all-ops, `uuidColumns` count vs feature-flag mismatch, `camelToSnake` leading-underscore + duplication, fragile `JSON.stringify` hash equality, resource leaks absolute paths, untested prompt/resource round-trips, `-32603` vs invalid-params, domain-* alias tool bloat, dead `knowledge.cli`/`knowledge.mcp` config fields, missing `requireAdmin` JSDoc, monolithic test.

### ✅ Confirmed good
- **#1341 feedback implemented well:** `capabilities` correctly declares `{ tools, prompts, resources }` (no tools regression — verified); README no longer says "Codex-compatible frontmatter"; the `smrt-review` tool/`smrt-code-review` skill name collision is resolved; resource URIs are well-formed (`smrt://knowledge/...`).
- **Security default posture holds:** route off by default, `requireAdmin` defaults true, `isKnowledgeAdmin`/`rolesFrom` are fail-closed (no prototype-pollution / truthy-coercion holes), dev-bypass relies on build-frozen `dev`, no path traversal. Risk is documented in standards.md + both templates' AGENTS.md.
- **Packaging verified:** `./knowledge` export + types re-exports + config wiring correct; `npm pack` ships `dist/knowledge.*` + `smrt-knowledge.json`; both templates' `CLAUDE.md` are the exact `@AGENTS.md` shim; `node scripts/check-standards.mjs` → 0 violations.

Most important single fix: regenerate-or-gitignore the stale artifact (it breaks your own gate), then fix the prompt-argument contract so the resources/prompts feature actually works from a generic harness.

packages/core/src/vite-plugin/sveltekit-generator.ts
```diff
@@ -1640,6 +1696,98 @@ ${optionsLoad}  const result = ${buildActionInvocationExpression(
...
+    INCLUDE_PROMPTS_BY_DEFAULT &&
+    url.searchParams.get('includePrompts') === 'true';
+
+  return json(sanitizeKnowledgeArtifact(artifact, {
+    includeDocs,
+    includePrompts,
+  }));
+};
+
+function readKnowledgeArtifact(): Record<string, any> {
```
willgriffin, line 1769: **🟠 Major (reliability + DoS) — reading the artifact from `process.cwd()` per request is fragile and unbounded.**

`resolve(process.cwd(), '.smrt/smrt-knowledge.json' | 'dist/smrt-knowledge.json')`:
- **cwd is unreliable across adapters.** Correct for adapter-node run from the project root, but Vercel/Cloudflare/Netlify functions execute from a bundled dir where neither path exists → the route throws `404` in production (confusing, not a leak — but the feature silently doesn't work). Static adapter has no server at all. Resolve relative to the route module (`import.meta.url`) or a `$lib`-bundled import, or document that only adapter-node is supported.
- **Blocking `readFileSync` + `JSON.parse` on every request**, no caching. The artifact can be large; combined with `requireAdmin:false` this is an unauthenticated DoS amplifier (repeated large-JSON parse + blocked event loop). Read + parse once at module load and cache.

(No path-traversal — paths are constants. Confirmed.)

packages/core/src/vite-plugin/sveltekit-generator.ts
```diff
@@ -1640,6 +1696,98 @@ ${optionsLoad}  const result = ${buildActionInvocationExpression(
...
+  }
+
+  const artifact = readKnowledgeArtifact();
+  const includeDocs =
+    INCLUDE_DOCS_BY_DEFAULT && url.searchParams.get('includeDocs') === 'true';
+  const includePrompts =
+    INCLUDE_PROMPTS_BY_DEFAULT &&
+    url.searchParams.get('includePrompts') === 'true';
+
+  return json(sanitizeKnowledgeArtifact(artifact, {
```
willgriffin, line 1762: **⚪ Minor (security) — no `Cache-Control` on the response.**

The generated route sets no cache headers, relying on SvelteKit defaults. For an admin-only endpoint that's usually fine, but with `requireAdmin:false` behind a shared CDN a sensitive JSON response could be cached and served to other clients. Emit `setHeaders({ 'cache-control': 'private, no-store' })` so it's safe regardless of CDN config. (No CORS headers are added → same-origin only by default, good.)

packages/types/src/knowledge.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export interface DomainKnowledgeConfig {
+  enabled?: boolean;
+  api?: {
+    enabled?: boolean;
+    basePath?: string;
+    includeDocs?: boolean;
+    includePrompts?: boolean;
+    requireAdmin?: boolean;
```
willgriffin, line 29: **⚪ Minor — `requireAdmin` has no JSDoc, despite being the single switch that makes the route public.**

A developer reading only the type sees no warning that `requireAdmin: false` exposes the full schema + endpoint map + dependency versions to anonymous users. Add a JSDoc comment here (and on `api.enabled`) flagging the exposure, mirroring the warnings in standards.md / the template AGENTS.md.

packages/core/src/manifest/smrt-knowledge.json
```diff
@@ -0,0 +1,330 @@
+{
+  "schemaVersion": 1,
+  "generatedAt": "2026-06-03T18:30:44.276Z",
+  "packageName": "@happyvertical/smrt-core",
+  "packageVersion": "0.27.0",
```
willgriffin: **🟠 Major — this committed artifact is already stale and breaks the PR's own `knowledge:check --strict` gate.**

`packageVersion` here is `0.27.0` but `packages/core/package.json` is now `0.27.1` (the branch was merged onto the v0.27.1 release without regenerating). `checkDomainKnowledgeArtifact` (`smrt-dev-mcp/src/knowledge/index.ts:452-478`) raw-hashes `package.json` and compares to the stored `sourceHashes.packageJson`, emitting `stale-domain-knowledge` (error under `--strict`). So `pnpm knowledge:check --strict` — listed in this PR's own validation — now fails on this branch.

This isn't a one-off: changesets bumps the version on **every** release, and the version is part of the hashed `package.json`, so the committed artifact re-goes-stale every release. Recommended fix: **gitignore `**/smrt-knowledge.json` like its sibling `static-manifest.json`** (already gitignored) and rely on the `dist/` build copy + the checker's `src/manifest/`→`dist/` fallback. If it must stay committed, exclude `version` from the package.json hash and regenerate it to `0.27.1` in this PR.

packages/core/src/knowledge.ts
```diff
@@ -0,0 +1,383 @@
...
+  };
+  const objects = Object.values(options.manifest.objects)
+    .filter((object) => object.decoratorConfig?.knowledge !== false)
+    .map((object) => buildKnowledgeObject(object));
+  const surfaces = objects.flatMap((object) => object.surfaces);
+  const manifestJson = JSON.stringify(options.manifest, null, 2);
+
+  return {
+    schemaVersion: 1,
+    generatedAt: new Date().toISOString(),
```
willgriffin, line 81: **⚪ Minor (determinism) — `generatedAt: new Date().toISOString()` churns a committed file on every build/hot-reload.**

The vite write path (`vite-plugin/index.ts` `writeDomainKnowledgeArtifact`) writes a fresh timestamp every time, with none of the `preserveGeneratedAtIfUnchanged` logic that `generate-manifest.js` has. So in a dev session every file change rewrites `.smrt/smrt-knowledge.json` with a new `generatedAt` even when nothing semantic changed — noisy git status, and the freshness layer can't distinguish "regenerated, identical" from "changed". If the artifact gets gitignored (see the artifact-staleness comment) this is moot for committed files; otherwise apply the preserve-guard on the vite path too.

packages/core/src/knowledge.ts
```diff
@@ -0,0 +1,383 @@
...
+  const allDependencies = {
+    ...record(packageJson.dependencies),
+    ...record(packageJson.devDependencies),
+    ...record(packageJson.peerDependencies),
+  };
+  const objects = Object.values(options.manifest.objects)
+    .filter((object) => object.decoratorConfig?.knowledge !== false)
+    .map((object) => buildKnowledgeObject(object));
+  const surfaces = objects.flatMap((object) => object.surfaces);
+  const manifestJson = JSON.stringify(options.manifest, null, 2);
```
willgriffin: **⚪ Minor — the manifest `sourceHash` is computed from the in-memory object, but the freshness checker hashes the on-disk file.**

Here the hash source is `JSON.stringify(options.manifest, null, 2)` (in-memory scanner manifest), while `checkDomainKnowledgeArtifact` recomputes it via `hashJsonFile` (read file → parse → re-stringify). These agree only if the manifest file is byte-stable. The scanner manifest (`static-manifest.json`) carries a `timestamp` field that changes every scan, so any manifest rescan that doesn't also regenerate `smrt-knowledge.json` flips this hash and reports `stale-domain-knowledge` even with zero semantic change. Normalize away volatile fields (`timestamp`) before hashing on both sides, or hash the written file rather than the in-memory object.

packages/core/src/knowledge.ts
```diff
@@ -0,0 +1,383 @@
...
+}
+
+function configuredOperations(config: unknown): string[] {
+  if (config === true) return [...STANDARD_OPERATIONS];
+  if (!config || typeof config !== 'object' || Array.isArray(config)) {
+    return [];
+  }
+  const recordConfig = config as { include?: string[]; exclude?: string[] };
+  const base =
+    Array.isArray(recordConfig.include) && recordConfig.include.length > 0
```
willgriffin: **⚪ Minor — `include: []` is treated as "all operations", the opposite of intent.**

`Array.isArray(recordConfig.include) && recordConfig.include.length > 0 ? include : STANDARD_OPERATIONS` — an explicit empty `include` (meaning "expose nothing") falls back to all 5 standard ops, so a class configured to expose nothing is documented as exposing everything. The same pattern exists in `mcpOperations` (`smrt-dev-mcp/src/knowledge/index.ts`). Drop the `length > 0` guard so `[]` yields `[]`.

packages/core/src/knowledge.ts
```diff
@@ -0,0 +1,383 @@
...
+    junctionCollections: objects.filter((object) =>
+      object.relationshipFeatures.includes('SmrtJunction'),
+    ).length,
+    hierarchicalObjects: objects.filter((object) =>
+      object.relationshipFeatures.includes('SmrtHierarchical'),
+    ).length,
+    polymorphicAssociations: objects.filter((object) =>
+      object.relationshipFeatures.includes('SmrtPolymorphicAssociation'),
+    ).length,
+    uuidColumns: fields.filter((field) => field.columnType === 'UUID').length,
```
willgriffin: **⚪ Minor — `uuidColumns` summary count disagrees with the per-object `uuidColumns` feature flag.**

This counts `field.columnType === 'UUID'` over `objects.flatMap(o => o.fields)`, but `fields` only contains user-defined fields — the synthetic `id` primary key isn't there. Meanwhile `relationshipFeatures` (line ~257) detects UUID by scanning `object.schema.columns` directly, which *does* include `id`. So a package of UUID-PK objects with no UUID FKs reports `relationshipsV2.uuidColumns: 0` while every object lists `uuidColumns` in `relationshipFeatures`. Count from `schema.columns` in both places for consistency.

packages/core/src/knowledge.ts
```diff
@@ -0,0 +1,383 @@
...
+    typeof exportsField !== 'object' ||
+    exportsField === null ||
+    Array.isArray(exportsField)
+  ) {
+    return [];
+  }
+  return Object.keys(exportsField).sort();
+}
+
+function camelToSnake(value: string): string {
```
willgriffin, line 392: **⚪ Nit — `camelToSnake` emits a leading underscore for names starting with a capital, and is duplicated across packages.**

`'MetaType'.replace(/[A-Z]/g, m => '_'+m.toLowerCase())` → `_meta_type`, which matches no real column key. Current field-naming conventions (lowercase-first) avoid triggering it, so it's latent — but the identical function is copy-pasted in `smrt-dev-mcp/src/knowledge/index.ts`. Anchor at first char (`/(?<!^)[A-Z]/` or strip the leading `_`) and extract one shared util.

packages/core/scripts/generate-manifest.js
```diff
@@ -77,4 +107,27 @@ if (import.meta.url === `file://${process.argv[1]}`) {
...
 
+function preserveGeneratedAtIfUnchanged(path, nextKnowledge) {
+  if (!existsSync(path)) {
+    return nextKnowledge;
+  }
+
+  try {
+    const current = JSON.parse(readFileSync(path, 'utf8'));
+    if (
+      JSON.stringify(current.sourceHashes ?? {}) ===
```
willgriffin: **⚪ Minor — `generatedAt` preservation hinges on `JSON.stringify` equality, which is key-order-sensitive.**

`JSON.stringify(current.sourceHashes) === JSON.stringify(next.sourceHashes)` only holds if both objects have identical key insertion order. It's stable for the current single code path, but a future reorder/added key — or an artifact last written by the vite path (which has no preserve logic) — makes this compare unequal even when hashes are semantically identical, defeating the preservation. Compare per-key, or canonicalize: `JSON.stringify(obj, Object.keys(obj).sort())`.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -267,32 +347,79 @@ async function main() {
...
-      description:
-        'Use this procedure when reviewing downstream SMRT projects.',
-      messages: [
-        {
-          role: 'user' as const,
-          content: {
-            type: 'text' as const,
-            text: renderAgentSkillMarkdown(REVIEW_SKILL_NAME),
+    if (name === DOMAIN_CODE_REVIEW_PROMPT) {
+      const context = await buildReviewContext(request.params.arguments as any);
```
willgriffin: **🟠 Major — `domain-code-review`/`domain-architecture` prompts can't receive `changedFiles` from any spec-compliant client.**

`buildReviewContext(request.params.arguments as any)` expects `changedFiles: string[]`, but MCP's `GetPrompt` `arguments` is typed `Record<string, string>` (string values only). A compliant client sending `changedFiles` as a string makes `buildReviewContext` throw `changedFiles.map is not a function` (empirically reproduced against the running server). And `ListPrompts` (line 343) declares **no `arguments`** for any prompt, so a harness can't even discover that these are parameterizable. With no args it silently falls back to `getChangedFiles()` against the *server's* cwd — a surprising behavior.

This is the core of the #1341 "expose via prompts" feedback, so it should work end-to-end. Fix: (a) declare `arguments` (`name`/`description`/`required`) on each prompt in `ListPrompts`; (b) accept the string-typed values the spec allows and parse `changedFiles` (e.g. newline/comma-split) into an array inside the handler instead of `as any`-casting a `Record<string,string>` into a `string[]` shape.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -303,25 +430,72 @@ async function main() {
...
-      contents: [
-        {
-          uri,
-          mimeType: 'text/markdown',
-          text: renderAgentSkillMarkdown(REVIEW_SKILL_NAME),
-        },
-      ],
-    };
+    if (uri === KNOWLEDGE_PROJECT_URI) {
+      const index = await buildKnowledgeIndex();
```
willgriffin, line 541: **⚪ Minor — the `smrt://knowledge/project` resource leaks absolute filesystem paths.**

`JSON.stringify(await buildKnowledgeIndex())` dumps the raw index, whose `rootDir` and per-package `directory` are absolute (e.g. `/.../node_modules/.pnpm/@happyvertical+ai@.../...`), exposing the host layout and resolved pnpm structure. The tool path (`reflect-domain-knowledge`) already returns a relativized/summarized projection; the resource diverges by serializing the raw index. Serialize a path-sanitized projection (drop/relativize `rootDir`, `directory`, object `filePath`) so the resource matches the tool's safe shape. Local dev MCP → low severity, but gratuitous.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -267,32 +347,79 @@ async function main() {
...
+            content: {
+              type: 'text' as const,
+              text: context.promptBundle.contextMarkdown,
+            },
+          },
+        ],
+      };
+    }
+
+    throw new Error(`Unknown prompt: ${name}`);
```
willgriffin: **⚪ Nit — unknown prompt/resource surfaces as `-32603` (Internal error).**

`throw new Error(\`Unknown prompt: ...\`)` (and the equivalent for resources) gets wrapped by the SDK as Internal Error. These are bad-params/not-found conditions; throwing `new McpError(ErrorCode.InvalidParams, ...)` (`-32602`) is more correct and friendlier to harnesses. No functional/security impact.

packages/smrt-dev-mcp/src/index.ts
```diff
@@ -377,6 +582,14 @@ async function main() {
           );
           break;
 
+        case 'check-domain-knowledge':
```
willgriffin, line 663: **⚪ Nit — `check-domain-knowledge` / `build-domain-*` tools are thin aliases that double the tool surface.**

These bodies are identical to their non-domain counterparts (the README even calls them aliases) and differ only by accepting `scope`/`package`, which the underlying functions already read off `args`. For harnesses that enumerate tools this is surface bloat. Consider folding `scope`/`package` into the existing tools rather than shipping parallel names.

packages/smrt-dev-mcp/src/mcp-stdio.test.ts
```diff
@@ -124,8 +167,12 @@ describe('smrt-dev-mcp stdio server', () => {
     expect(skill.referenceFiles[0].content).toContain('SMRT Review Output');
 
     const promptsResult = await client.listPrompts();
-    expect(promptsResult.prompts.map((prompt) => prompt.name)).toContain(
-      'smrt-code-review',
+    expect(promptsResult.prompts.map((prompt) => prompt.name)).toEqual(
+      expect.arrayContaining([
+        'smrt-code-review',
+        'domain-code-review',
```
willgriffin, line 173: **⚪ Minor — the new prompt/resource round-trips are listed but never exercised.**

The smoke test now *lists* `domain-code-review`/`domain-architecture` and `smrt://knowledge/project`, but it only ever `getPrompt`/`readResource`s the pre-existing `smrt-code-review` skill + agent-skills resource. The new `getPrompt({name:'domain-code-review'})` and `readResource('smrt://knowledge/project' | '.../package/...')` paths — all the code this PR adds — are untested, which is exactly why the prompt-argument bug above slipped through. Add Get/Read assertions for the new prompts/resources, including the string-arg case and an unknown-package case.

packages/types/src/knowledge.ts
```diff
@@ -0,0 +1,107 @@
...
+  };
+  cli?: boolean;
```
willgriffin: **⚪ Minor — `knowledge.cli` and `knowledge.mcp` are dead config fields.**

Both templates set `cli: true, mcp: true` and standards.md says "CLI and MCP exposure are on by default", but no reader consumes `config.knowledge.cli`/`.mcp` anywhere (the `smrt dev:knowledge-*` CLI and the dev-mcp tools don't gate on them). So the templates advertise toggles that do nothing. Either wire them into the CLI/MCP gating or drop them from the type + templates + standards doc to avoid a misleading contract. (The `config.cli`/`config.mcp` you'll find elsewhere are the unrelated `@smrt()` decorator flags.)

packages/core/src/knowledge.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      join(rootDir, 'src', 'prompts', 'review.ts'),
+      "export const review = definePrompt('orders.review', {});\n",
+    );
+  });
+
+  afterEach(() => {
+    rmSync(rootDir, { recursive: true, force: true });
+  });
+
+  it('builds a domain knowledge artifact from a SMRT manifest', () => {
```
willgriffin: **⚪ Nit — the artifact test is one large `it` with many assertions.**

`builds a domain knowledge artifact ...` bundles many distinct checks; splitting into focused `it`s (surfaces, relationship counts, doc/prompt inclusion, exclusion via `knowledge:false`) would improve failure isolation when one assertion regresses.

packages/core/src/manifest/smrt-knowledge.json
```diff
@@ -0,0 +1,330 @@
...
+  "sourceManifestPath": "src/manifest/static-manifest.json",
```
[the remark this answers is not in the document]
willgriffin: Fixed in 6be49ffdd. The source smrt-knowledge artifact is no longer tracked, generated artifacts are ignored, and the core package build artifact now points at the built dist manifest path rather than requiring an ignored source manifest in a fresh checkout.

packages/core/src/vite-plugin/sveltekit-generator.ts
```diff
@@ -1640,6 +1696,98 @@ ${optionsLoad}  const result = ${buildActionInvocationExpression(
...
+  if (!dev && REQUIRE_ADMIN && !isKnowledgeAdmin(locals as Record<string, any>)) {
+    throw error(403, 'SMRT knowledge requires dev mode or admin access');
+  }
+
+  const artifact = readKnowledgeArtifact();
+  const includeDocs =
+    INCLUDE_DOCS_BY_DEFAULT && url.searchParams.get('includeDocs') === 'true';
+  const includePrompts =
+    INCLUDE_PROMPTS_BY_DEFAULT &&
+    url.searchParams.get('includePrompts') === 'true';
```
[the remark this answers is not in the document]
willgriffin: Fixed in 6be49ffdd. The generated route now treats config booleans as defaults and lets includeDocs/includePrompts query params override those defaults explicitly.

packages/core/src/knowledge.ts
```diff
@@ -0,0 +1,383 @@
...
+    columnType: columnType(object, name),
+  }));
+  const relationships = fields
+    .filter((field) => RELATIONSHIP_FIELD_TYPES.has(field.type))
+    .map((field) => ({
+      name: field.name,
+      type: field.type,
+      related: field.related,
+      columnType: field.columnType,
+    }));
```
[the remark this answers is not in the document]
willgriffin: Fixed in 6be49ffdd. Relationship entries now preserve the field required flag from the manifest, with focused coverage in packages/core/src/knowledge.test.ts.

packages/core/scripts/generate-manifest.js
```diff
@@ -77,4 +107,27 @@ if (import.meta.url === `file://${process.argv[1]}`) {
...
+        ...nextKnowledge,
+        generatedAt: current.generatedAt,
+      };
+    }
+  } catch {
+    // Replace malformed artifacts.
+  }
+
+  return nextKnowledge;
+}
```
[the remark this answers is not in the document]
willgriffin: Fixed in 6be49ffdd. generatedAt preservation now compares a stable semantic artifact with only generatedAt removed, so prompt/surface/metadata changes advance the timestamp.

packages/core/src/vite-plugin/sveltekit-generator.ts
willgriffin: Fixed in 6be49ffdd. Public knowledge responses now strip dependencies, source hashes, source manifest path, and agent doc path in addition to docs/prompts, and generation warns when requireAdmin is disabled.

packages/core/src/vite-plugin/sveltekit-generator.ts
willgriffin: Fixed in 6be49ffdd. The generated route embeds the artifact at generation time as a module constant, removing process.cwd(), per-request filesystem reads, and per-request JSON parsing.

packages/core/src/vite-plugin/sveltekit-generator.ts
willgriffin: Fixed in 6be49ffdd. The generated knowledge route now sets Cache-Control: private, no-store.

packages/types/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. Added explicit JSDoc for api.enabled and api.requireAdmin, including the public-exposure warning for requireAdmin: false.

packages/core/src/manifest/smrt-knowledge.json
willgriffin: Fixed in 6be49ffdd. The source artifact is removed from git and smrt-knowledge.json is ignored; package build still emits dist/smrt-knowledge.json and export validation covers the published artifact shape.

packages/core/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. The Vite write path now preserves generatedAt when the stable semantic artifact is unchanged, and the generated source artifact is ignored instead of committed.

packages/core/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. The builder and freshness checker both hash canonical manifest JSON with the top-level timestamp removed, so generated timestamps do not stale otherwise identical artifacts.

packages/core/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. configuredOperations now distinguishes an explicit include: [] from an omitted include and returns no operations for the explicit empty allowlist.

packages/core/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. uuidColumns summary now counts schema.columns UUID entries, matching the per-object relationship feature detection.

packages/core/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. The camel-to-snake helper no longer emits a leading underscore for capitalized names in both affected packages. I kept the two local copies rather than adding cross-package coupling for this small helper.

packages/core/scripts/generate-manifest.js
willgriffin: Fixed in 6be49ffdd. generatedAt preservation now uses stable sorted JSON for the semantic artifact comparison instead of raw JSON.stringify key order.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Fixed in 6be49ffdd. The domain prompts now declare arguments and parse MCP string arguments, including changedFiles as JSON, newline, or comma-separated input, with stdio coverage.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Fixed in 6be49ffdd. Project/package resources now serialize sanitized projections: rootDir is normalized to ., package directory is omitted, and absolute object file paths are redacted.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Fixed in 6be49ffdd. Unknown prompt/resource/package paths now throw McpError(ErrorCode.InvalidParams, ...) instead of generic Error.

packages/smrt-dev-mcp/src/index.ts
willgriffin: Kept intentionally in 6be49ffdd. The domain-named tools are aliases, but they are part of the accepted implementation plan for downstream domain knowledge surfaces. The aliases continue to share implementation with the base tools, so behavior remains consolidated while preserving the requested API shape.

packages/smrt-dev-mcp/src/mcp-stdio.test.ts
willgriffin: Fixed in 6be49ffdd. The MCP stdio smoke test now exercises domain-code-review and domain-architecture getPrompt calls, project/package resource reads, string changedFiles parsing, and unknown-package rejection.

packages/types/src/knowledge.ts
willgriffin: Fixed in 6be49ffdd. Removed knowledge.cli and knowledge.mcp from the shared contract, Vite defaults, templates, and standards wording so the config no longer advertises dead toggles.

packages/core/src/knowledge.test.ts
willgriffin: Fixed in 6be49ffdd. Split the domain knowledge artifact test into focused cases for metadata/source hashes, surfaces/prompts, relationship summaries, and exclusion behavior.

PR #1440, opened 2026-06-07
fix(smrt-svelte): link form feedback to inputs

web3blind:
## Summary

This PR hardens the `smrt-svelte` form primitives from #1420 by making their validation/help/status text available to assistive technology.

It updates the core text, textarea, select, number, and money inputs so that:

- dynamic supporting or validation text receives a stable generated id;
- the related control points at that text with `aria-describedby` only when the text is present;
- invalid text/validation states set `aria-invalid="true"`;
- changing validation/status text is announced politely without moving focus.

I also added a focused regression test that pins those relationships across the affected form primitives.

## Validation

- `python3` static assertions over the affected Svelte sources passed locally.
- `git diff --check` passed.
- `pnpm --filter @happyvertical/smrt-svelte test -- src/components/forms/__tests__/accessibility-relationships.test.ts` could not run in my environment because the workspace install requires private `@happyvertical/*` GitHub Packages access and the local Node version is v22 while the repo requires Node >=24.

Closes #1420 (partial component-family pass).


packages/smrt-svelte/src/components/forms/NumberInput.svelte
```diff
@@ -217,7 +220,7 @@ function handleInput(e: Event) {
   </div>
 
   {#if showInvalid}
-    <div class="validation-error">
+    <div id={validationErrorId} class="validation-error" aria-live="polite">
```
willgriffin, line 223: This live region is created only after `showInvalid` becomes true, so the first validation message may not be announced reliably by screen readers. Live/status regions need to exist before their contents change. Please keep an empty error/status container mounted (for example `role="status"`/`aria-live="polite"`) and update its contents when validation changes.

packages/smrt-svelte/src/components/forms/MoneyInput.svelte
```diff
@@ -310,9 +313,9 @@ function handleBlur() {
   <input type="hidden" name="{name}_cents" value={value ?? ''} />
 
   {#if error}
-    <div class="validation-error">{error}</div>
+    <div id={validationErrorId} class="validation-error" aria-live="polite">{error}</div>
```
willgriffin, line 316: Same live-region issue here: adding the `aria-live` node at the same time as the error text makes the announcement unreliable. Keep the live region in the DOM ahead of time and only change its contents when `error`/range validation changes.

packages/smrt-svelte/src/components/forms/__tests__/accessibility-relationships.test.ts
```diff
@@ -0,0 +1,57 @@
+import { readFileSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+const formsDir = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+
+function source(component: string) {
```
willgriffin, line 8: These regression tests only inspect source strings, so they can pass even if Svelte renders the wrong DOM or the state transition never exposes the relationship. Please mount the components in jsdom and assert the actual `aria-describedby` target exists, the referenced text is present, and `aria-invalid` toggles for invalid states.

willgriffin (commented):
One PR-description issue that cannot be anchored inline: this says `Closes #1420`, but #1420 also covers the remaining form primitives, `FormGroup`, `Form` live status, and axe coverage. Since this PR is a partial component-family pass, please change that to `Refs #1420` unless the remaining acceptance criteria are handled here.

PR #1574, opened 2026-06-22
Publish SMRT packages to npmjs

willgriffin:
Refs #1563.

## Summary
- switch SMRT package publish targets from GitHub Packages to public npmjs
- wire release and publish dry-run workflows to use NPM_TOKEN while keeping GitHub Packages auth for private install dependencies
- update publish standards/docs and generated publishConfig helper

## Self-review
No blocking findings. The remaining GitHub Packages references are scoped to dependency installation for existing @happyvertical SDK packages; package publishConfig and publish jobs now target https://registry.npmjs.org.

## Validation
- pnpm install --frozen-lockfile --loglevel error
- pnpm run check:standards
- pnpm run build
- node scripts/validate-publish-packages.js
- actionlint .github/workflows/publish.yml .github/workflows/publish-dry-run.yml .github/workflows/on-merge-main.yml
- ruby -e 'require "yaml"; ARGV.each { |file| YAML.load_file(file) }' .github/actions/setup-environment/action.yml
- git diff --check
- pre-push hooks
- review follow-up: actionlint .github/workflows/publish.yml .github/workflows/publish-dry-run.yml
- review follow-up: git diff --check
- review follow-up: pre-commit and pre-push hooks

## Note
Anonymous install of SMRT packages from npmjs may still require publishing non-SMRT @happyvertical SDK dependencies, such as @happyvertical/sql and @happyvertical/utils, to npmjs as well. Those are outside this SMRT issue scope.


.github/workflows/publish.yml
```diff
@@ -56,16 +56,17 @@ jobs:
...
+          registry-url: 'https://registry.npmjs.org'
+          auth-token: ${{ secrets.NPM_TOKEN }}
+          github-packages-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Prepare versioned release workspace
         id: prepare
         env:
           CI: true
           GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
willgriffin: Minor inconsistency: this prepare step keeps `NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` while flipping `NPM_TOKEN` to `secrets.NPM_TOKEN`. Every other step in the file moved *both* to `NPM_TOKEN` (e.g. lines 237–238, 315–316). It's harmless today because this step (changeset version + build) doesn't authenticate to a registry, but it's a latent footgun: if a future change here ever touches npmjs via `NODE_AUTH_TOKEN`, it'll use the wrong (GitHub) token. Either align both to `NPM_TOKEN`, or add a one-line comment explaining why the GHP token is intentionally retained here. (Same pattern in `publish-dry-run.yml:43`.)

.github/workflows/publish.yml
```diff
@@ -281,14 +283,24 @@ jobs:
       - name: Extract versioned workspace and build outputs
         run: tar -xf release-artifacts/publish-release-workspace.tar
 
+      - name: Verify npm publish token
```
willgriffin, line 295: Nice fail-safe, but it runs only in `publish-release` — after `prepare-release` and the four `validate-release-shards` have already completed (~40 min of CI). If `NPM_TOKEN` is unset, the whole pipeline burns before failing here. No half-published state results (the commit/push/tag happen in a later step), so this is low priority, but moving the presence check into `prepare-release` (or a tiny preflight job) would fail fast on a missing secret.

willgriffin (commented):
## Review: Publish SMRT packages to npmjs (#1574)

The mechanics of the registry switch are sound. I verified the parts that are easy to get wrong, and they're correct:

- **All publishable packages** now pin `publishConfig.registry = https://registry.npmjs.org` (including the three that previously had none: `app-cli`, `template-sveltekit`, `template-site-static-json`). Since `publishConfig.registry` takes precedence over the `@happyvertical:registry=https://npm.pkg.github.com` scope routing in the repo-root `.npmrc`, `changeset publish` will correctly target npmjs. ✅
- **`packages: write` → `packages: read`** is correct least-privilege: publishing no longer needs GH Packages write, and the read scope is retained so the new "Configure GitHub Packages auth" step can still install the `@happyvertical/*` SDK deps from GHP. Both caller (`on-merge-main.yml`) and callee (`publish.yml`) are updated, and `secrets: inherit` passes `NPM_TOKEN` through. ✅
- The new `~/.npmrc` GHP block coexists cleanly with the npmjs default-registry auth (GHP auth baked as `GITHUB_TOKEN`, npmjs as `NPM_TOKEN`). ✅
- `deploy-docs` dropping `auth-token` is safe — `docs/package.json` has no `@happyvertical` deps. ✅
- No stale "publish to GitHub Packages" prose left (CHANGESETS.md, standards.md, release notes all updated). ✅

### 🔴 Main concern: `Closes #1563` will auto-close an issue whose acceptance test still fails

Issue #1563's acceptance criterion is that **`npm install @happyvertical/smrt-core` succeeds on a clean machine with no auth token**. This PR does not achieve that yet: the published smrt packages declare *runtime* (`dependencies`) on non-smrt SDK packages that are **only on GitHub Packages**, e.g. `packages/core/package.json` depends on `@happyvertical/ai`, `@happyvertical/sql`, `@happyvertical/files`, `@happyvertical/json`, `@happyvertical/logger`, `@happyvertical/utils` (all `catalog:`). On a clean machine, npm will try to resolve those transitive deps from npmjs, 404, and the install fails.

Nothing in the pipeline catches this: `validate-publish-packages.js` uses `npm pack --dry-run` (doesn't resolve deps), and `npm publish` doesn't verify dep availability — so the release will silently ship packages that can't be anonymously installed.

The PR's "Note" already acknowledges the SDK packages need to land on npmjs too, and treats that as out of scope — which is reasonable to do separately. The problem is purely the `Closes #1563` keyword: merging will auto-close the issue even though its acceptance test still fails. **Recommend changing `Closes #1563` → `Refs #1563`** (and keeping it open / tracking the SDK-publish follow-up), so the issue closes only once anonymous install actually works.

The two inline comments below are minor.

.github/workflows/publish.yml
willgriffin: Aligned `NODE_AUTH_TOKEN` to `secrets.NPM_TOKEN` in both `publish.yml` and `publish-dry-run.yml` prepare steps in `9a2db2864`.

.github/workflows/publish.yml
willgriffin: Added the same `NPM_TOKEN` presence check to the `prepare-release` job immediately after checkout in `9a2db2864`, so a missing npm token fails before setup/build/sharded validation. I left the later publish-job check in place as a defensive guard.

PR #1575, opened 2026-06-22
[codex] Reuse request-scoped auth and entitlement work

willgriffin:
## Summary

Fixes #1570.
Fixes #1573.

- Reuse the active `Membership` already loaded for a request/session when resolving user permissions.
- Carry `membership` through `SessionContext`, SvelteKit locals, request permission context, and resource-list session resolution.
- Batch subscription threshold usage summaries by usage window, with a back-compatible single-metric fallback.
- Add `SubscriptionResolver.create()` and `EntitlementResolutionContext` so request-scoped callers can reuse default readers plus preloaded subscription/plan state.
- Add collection-level and meter-level batch usage APIs for persisted tenant metrics and tenant-scoped `ai.*` metrics.

## Root Cause

The default SaaS request path repeatedly re-resolved the same tenant membership and subscription entitlement state. Subscription threshold evaluation also fanned out to one usage summary call per threshold, which scaled with threshold count even when all thresholds shared the same usage window.

## Review Notes

I did a manual pass over request-context propagation, tenancy boundaries, public exports, generated SQL columns, and the reusable entitlement context API. One defect was found during review: a caller-provided context plan without an `id` could be trusted without proving it matched `subscription.planId`. That now fails closed and has regression coverage.

SMRT deterministic review context raised only routing warnings for public entrypoint/docs/generated-surface changes; those were checked manually.

## Validation

- `pnpm --filter @happyvertical/smrt-users test`
- `pnpm --filter @happyvertical/smrt-users typecheck`
- `pnpm --filter @happyvertical/smrt-users build`
- `pnpm --filter @happyvertical/smrt-subscriptions test`
- `pnpm --filter @happyvertical/smrt-subscriptions typecheck`
- `pnpm --filter @happyvertical/smrt-subscriptions build`
- `pnpm exec biome check --max-diagnostics 200 <changed files>`
- `git diff --check`
- pre-commit and pre-push hooks

Biome reports existing non-null assertion warnings in older users tests and resolver code; no blocking errors were introduced. Pre-push workflow validation also reports an existing line-length warning in `.github/workflows/test-suite.yml`, while still passing.

### DIFF
--- a/packages/subscriptions/src/services/subscription-resolver.ts
+++ b/packages/subscriptions/src/services/subscription-resolver.ts
@@ -1,8 +1,12 @@
+import { SubscriptionPlanCollection } from '../collections/SubscriptionPlanCollection.js';
+import { TenantSubscriptionCollection } from '../collections/TenantSubscriptionCollection.js';
 import type { SubscriptionPlan } from '../models/SubscriptionPlan.js';
 import type { TenantSubscription } from '../models/TenantSubscription.js';
 import type {
   EntitlementResolution,
+  EntitlementResolutionContext,
   PlanThreshold,
+  SmrtClassOptions,
   Subscriber,
   SubscriptionResolverOptions,
   ThresholdEvaluation,
@@ -15,6 +19,7 @@ import {
   isValidThreshold,
 } from '../utils.js';
 import { evaluateThreshold } from './threshold-evaluator.js';
+import { TenantUsageMeter } from './usage-meter.js';
 
 export interface SubscriptionPlanReader {
   get(criteria: { id: string }): Promise<SubscriptionPlan | null>;
@@ -77,6 +82,41 @@ export class SubscriptionResolver {
   constructor(private readonly readers: SubscriptionResolverReaders) {}
 
   /**
+   * Build the default resolver readers once for a request or app lifecycle and
+   * reuse the returned resolver across entitlement checks.
+   */
+  static async create(
+    classOptions: SmrtClassOptions = {},
+  ): Promise<SubscriptionResolver> {
+    const [plans, subscriptions, usage] = await Promise.all([
+      SubscriptionPlanCollection.create(classOptions),
+      TenantSubscriptionCollection.create(classOptions),
+      TenantUsageMeter.create(classOptions),
+    ]);
+    return new SubscriptionResolver({ plans, subscriptions, usage });
+  }
+
+  /**
+   * Load the subscription/plan pair used by entitlement resolution. Callers
+   * that need both the entitlement snapshot and the backing records can load
+   * this once, then pass it back via `options.context`.
+   */
+  async loadEntitlementContext(
+    subscriberOrTenantId: Subscriber | string,
+    options: SubscriptionResolverOptions = {},
+  ): Promise<EntitlementResolutionContext> {
+    const subscriber = toSubscriber(subscriberOrTenantId);
+    const now = options.now ?? new Date();
+    const subscription = await this.resolveSubscription(
+      subscriber,
+      now,
+      options.context,
+    );
+    const plan = await this.resolvePlan(subscription, options.context);
+    return { subscription, plan };
+  }
+
+  /**
    * Polymorphic entitlement resolution. Works for both `'tenant'`-kind and
    * `'external'`-kind subscribers and is the preferred surface — the
    * `resolveTenantEntitlements(tenantId)` method below is a thin wrapper.
@@ -86,15 +126,14 @@ export class SubscriptionResolver {
     options: SubscriptionResolverOptions = {},
   ): Promise<EntitlementResolution> {
     const now = options.now ?? new Date();
-    const subscription = await this.findCurrentSubscription(subscriber, now);
+    const { subscription, plan } = await this.loadEntitlementContext(
+      subscriber,
+      { ...options, now },
+    );
 
     if (!subscription) {
       return emptyResolution(subscriber);
     }
-
-    const plan = subscription.planId
-      ? await this.readers.plans.get({ id: subscription.planId })
-      : null;
 
     if (!plan || !plan.isActive() || !subscription.isEntitled(now)) {
       return {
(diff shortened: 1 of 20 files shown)


packages/subscriptions/src/services/subscription-resolver.ts
```diff
@@ -226,4 +371,79 @@ function emptyResolution(subscriber: Subscriber): EntitlementResolution {
   };
 }
 
+function hasContextValue<K extends keyof EntitlementResolutionContext>(
+  context: EntitlementResolutionContext | undefined,
+  key: K,
+): boolean {
+  return context !== undefined && Object.hasOwn(context, key);
```
willgriffin: Corroborating the automated reviewer here, with the concrete impact and a suggested fix.

`hasContextValue` keys off `Object.hasOwn`, so a context key set to an **explicit `undefined`** is treated as "present" and then coerced via `?? null`. The `EntitlementResolutionContext` fields are optional (`T | null | undefined`), and the doc comments define `null` = "already resolved as absent" but leave `undefined` = "not provided → fetch". `Object.hasOwn` collapses that distinction.

Concrete footgun — a request-scoped caller assembling a partial context:
```ts
const ctx = { subscription, plan: maybePlan }; // maybePlan: SubscriptionPlan | undefined
await resolver.resolveTenantEntitlements(subscriber, { context: ctx });
```
When `maybePlan` is `undefined`, `hasContextValue(ctx, 'plan')` is `true` → `plan = null` → `resolveTenantEntitlements` returns a **closed entitlement** (plan treated as absent) instead of loading the real plan from the DB. Same hazard for `subscription`. Object spreads and destructured optionals make an explicit-`undefined` easy to produce by accident.

Suggest gating on the value rather than key presence (keeps `null` meaning "resolved absent"):
```ts
function hasContextValue<K extends keyof EntitlementResolutionContext>(
  context: EntitlementResolutionContext | undefined,
  key: K,
): boolean {
  return context?.[key] !== undefined;
}
```
…or, if present-with-`undefined` is meant to be valid, document on `EntitlementResolutionContext` that callers must **omit** a key (never set it to `undefined`) to request a fresh read.

willgriffin (commented):
## Review — request-scoped auth & entitlement reuse

Independent pass over request-context propagation, the entitlement-context reuse + fail-closed validation, the batched usage path (SQL param ordering, window grouping, AI-vs-persisted routing), and the public-surface changes. The core design is sound and well-covered by tests. In particular:

- The batched threshold path correctly mirrors the single-path routing (`ai.*` → `_smrt_ai_usage` for tenant subscribers, everything else → persisted metrics), so moving every real `TenantUsageMeter` caller onto the batch path does not change results.
- `summarizeUsageBatch` SQL param ordering — including the `params.splice(2, 0, externalId)` for external subscribers — and the `window_start < end AND window_end > start` overlap predicate match the existing `summarizeUsage`.
- `Promise.all` over window groups writes `evaluations[index]` on disjoint indices, so no race and no holes.
- The fail-closed context checks (`assertPlanMatchesSubscription` requiring `plan.id === subscription.planId`, `assertSubscriptionMatchesSubscriber`) and the `membership.userId/tenantId` guard correctly prevent a caller-supplied context/membership from being trusted for the wrong subscriber.

Two items to look at (both non-blocking) are inline. Minor notes:

- **Asymmetric failure mode** on a context/identity mismatch: `SubscriptionResolver` *throws* (`assert*MatchesSubscriber`/`Subscription`), whereas `PermissionResolver` silently returns an empty deny result when `membership.userId/tenantId` don't match. Both are fail-closed and safe, but the inconsistency can surprise — and the silent-deny can mask a caller bug rather than surfacing it.
- **Nit:** `emptyUsageSummary` is duplicated in `subscription-resolver.ts` and `usage-meter.ts`. The resolver-side copy is only reachable for a custom `UsageSummaryReader` that under-returns; the bundled `TenantUsageMeter.summarizeBatch` always returns a full set keyed by every requested metric.

Note: CI was still mid-run at review time (Test/Typecheck/Coverage/Knowledge pending; Build + Lint green).

packages/users/src/sveltekit/resource-list-handler.ts
```diff
@@ -117,6 +117,7 @@ export interface CliResource {
 
 export interface ResolvedSession {
   user: SessionLocals['user'];
+  membership: SessionLocals['membership'];
```
willgriffin: `membership` is added here as a **required** property, and `ResolvedSession` is public API: `createCliResourceListHandler` accepts a caller-supplied `resolveSession?: (event) => Promise<ResolvedSession>` (≈ line 187). Any downstream consumer that provides its own `resolveSession` will now fail to typecheck until it also returns `membership`. Same applies to `SessionLocals` in `sveltekit/types.ts` (also now required).

Since the framework's own `defaultResolveSession` always populates it, making it optional costs nothing internally and avoids the break for custom resolvers:
```ts
membership?: SessionLocals['membership'];
```
If the required field is intentional, worth calling out for the ergot.io / anytown.ai consumers so their custom session resolvers (if any) get updated.

packages/subscriptions/src/services/subscription-resolver.ts
```diff
@@ -226,4 +371,79 @@ function emptyResolution(subscriber: Subscriber): EntitlementResolution {
...
+}
```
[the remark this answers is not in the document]
willgriffin: Fixed in `b41182432`: `hasContextValue()` now treats `undefined` as absent (`context?.[key] !== undefined`), so omitted/undefined context values fall back to the readers while explicit `null` remains the resolved-absent sentinel. Added regression coverage for `{ plan: undefined, subscription: undefined }` falling back to the DB readers.

packages/subscriptions/src/services/subscription-resolver.ts
willgriffin: Fixed in `b41182432`: the context helper now preserves the intended `undefined` vs `null` distinction, and the new regression test verifies undefined context values do not fail closed or skip reader lookups.

packages/users/src/sveltekit/resource-list-handler.ts
willgriffin: Fixed in `b41182432`: `membership` is now optional on the public `SessionContext`, `SessionLocals`, and `ResolvedSession` types. The framework defaults still populate `membership: null`, but custom downstream `resolveSession` implementations no longer need to return it to typecheck.

PR #1644, opened 2026-06-24
[codex] feat(reports): add materialized aggregate reports

willgriffin:
## Summary
- Adds `@happyvertical/smrt-reports` for decorator-defined materialized aggregate reports, refresh state, and schedule/job integration.
- Extends scanner/manifest support so report classes are discoverable and carry report metadata.
- Consumes `buildAggregate` / `bucketExpr` from `@happyvertical/sql` instead of keeping aggregate SQL generation in SMRT.

## Dependency
- Built on SDK PR happyvertical/sdk#1054. This SMRT PR should stay draft until that SDK PR is merged, released, and the SMRT catalog/lockfile is updated to the released `@happyvertical/sql` version containing `buildAggregate`.

## Validation
- SDK PR package linked locally for `@happyvertical/sql` while validating SMRT.
- `pnpm --filter @happyvertical/smrt-reports typecheck`
- `pnpm --filter @happyvertical/smrt-reports build`
- `pnpm --filter @happyvertical/smrt-reports test`
- `pnpm --filter @happyvertical/smrt-core exec vitest run src/scanner/__tests__/manifest-generator.test.ts`
- `pnpm --filter @happyvertical/smrt-scanner exec vitest run src/__tests__/manifest-adapter.test.ts src/__tests__/oxc-parser.test.ts`
- `pnpm knowledge:check --strict --format markdown`
- staged-file Biome check for changed TS/JSON files
- gitleaks staged scan
- package guardrail scripts and `pnpm check:package-dag`

## Known local gate note
- Full repo `biome ci .` currently fails on existing import-order diagnostics under `packages/smrt-ui/src/components/forms/*`, which this branch does not touch.


willgriffin (commented):
## Review — `feat(reports): add materialized aggregate reports`

Substantial, cleanly-structured new package. The spec compiler / refresh engine split matches the locked PRD design (#1633), and `refresh-integration.test.ts` is genuinely strong — real in-memory SQLite exercising incremental recompute, `avg`, empty-group deletes, tenant fan-out, schedules, the onChange interceptor, and the stateless queued task. Scanner/core wiring (`SmrtReport`/`SmrtReportCollection` added to both `FRAMEWORK_BASE_CLASSES` and `FRAMEWORK_ABSTRACT_BASE_NAMES`, `report` config plumbed through manifest → conflict-column derivation) follows the framework-abstract-base precedent correctly.

### Blocking (acknowledged in the PR body)
All red CI — Typecheck, Test Packages (1/3), and every Publish Dry Run shard — traces to one cause: the `@happyvertical/sql` catalog pin (`^0.74.7`) does not yet export `buildAggregate` / `bucketExpr` / `AggregateSpec` / `AggregateSelectExpr` / `AggregateTimeBucketUnit`. Those land in SDK PR happyvertical/sdk#1054. (`src/refresh.ts:315` `TS7006 'expr' implicitly any` is a cascade off the same missing types, not an independent bug.)

This PR **cannot merge to a green `main`** until sdk#1054 is merged + released and the catalog/lockfile are bumped to the version exporting `buildAggregate`. Recommend converting this PR to **draft** until then so it can't be merged accidentally, and adding a checklist item for the catalog bump. Everything below is independent of the SDK dependency.

### Findings
Inline comments cover, in rough priority order:
- Lock acquisition is not atomic (check-then-set TOCTOU) — `lock` reads as a correctness guarantee it doesn't enforce.
- `assertReportTablesReady` gap for `incremental` when `trackRuns:false && lock:false` (watermarks table still required).
- Incremental + time-bucket grouping is untested and depends on dialect-sensitive `HAVING`-on-alias semantics.
- `adapterTypeFromDb` `as any` reach-ins vs the "no private reach-ins" convention.
- `refreshIfStale` hardcodes `tenant_id`/`refreshed_at` and adds a per-read query.
- Test-convention alignment (mock DB in the unit test; `.spec.ts` vs `.test.ts`).
- Scheduler poll timer never `unref()`d.

### Minor
`README.md` still says "39 packages total" — adding `reports` makes 40.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,993 @@
...
+    }
+  }
+
+  const expiresAt = new Date(now.getTime() + ttlMs);
+  const id = stableReportId([
+    'lock',
+    definition.reportClassName,
+    scope.scopeKey,
+  ]);
+  await db.upsert(REPORT_LOCKS_TABLE, ['report_class', 'scope_key'], {
```
willgriffin: **Lock acquisition isn't atomic.** `acquireLock` does a `SELECT` (line 611) to check for a live lock, then a separate `db.upsert` here to claim it. Two refreshes for the same `(report_class, scope_key)` can both observe "no live lock" and both upsert — each gets a non-null handle and proceeds. Because the upsert is last-writer-wins on `owner_id`, only one `release()` will match (the other no-ops), but **mutual exclusion is never actually enforced**.

For incremental refresh this lets two runners read the same `watermarkBefore` and double-process the same window. The job queue + `tenantJobCap` may serialize most callers in practice, but the `lock` option reads as a hard guarantee it doesn't currently provide. Consider an atomic conditional claim — e.g. `INSERT ... ON CONFLICT DO UPDATE SET owner_id=excluded.owner_id, expires_at=... WHERE locks.owner_id IS NULL OR locks.expires_at < ?` and then verify the persisted `owner_id == ownerId` — instead of check-then-set.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,993 @@
...
+  release(): Promise<void>;
+}
+
+function adapterTypeFromDb(
+  db: DatabaseInterface,
+  fallback?: SqlAdapterType,
+): SqlAdapterType {
+  const candidate =
+    fallback ??
+    (db as any).type ??
```
willgriffin: Reaching into `(db as any).type` / `.config?.type` / `.options?.type` casts through the adapter's internals, which AGENTS.md calls out under **"No private reach-ins"**. The URL-sniff fallback also only recognizes sqlite/postgres, so a `duckdb`/`json` adapter that exposes neither a recognized `type` nor a `postgres://` URL will throw here even though those adapter types are otherwise supported. If `DatabaseInterface` has no public adapter-type accessor, adding one upstream in `@happyvertical/sql` would be cleaner than the `as any` probes.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,993 @@
...
+  const exists = await tableExists(options.db, reportTable);
+  if (!exists) {
+    throw new Error(
+      `Report table '${reportTable}' does not exist for ${reportCtor.name}. ` +
+        `Run smrt db:migrate before refreshing reports.`,
+    );
+  }
+
+  if (options.trackRuns !== false || options.lock !== false) {
+    await assertReportTablesReady(options.db);
```
willgriffin: `assertReportTablesReady` (runs/watermarks/locks) is skipped when `trackRuns === false && lock === false`. But `incrementalReport` always reads/writes `_smrt_report_watermarks` regardless of those flags — so an incremental refresh with both disabled and no watermarks table throws a raw SQL error instead of the friendly "Run smrt db:migrate before refreshing reports" guidance the assert is meant to provide. Consider always asserting the watermarks table when the resolved mode is `incremental`.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,993 @@
...
+  const groupingColumns = getActualGroupingColumns(definition);
+  const groupHaving = Object.fromEntries(
+    groupingColumns.map((column) => [column, groupRow[column]]),
+  );
+  const spec = withRefreshFilters(compileReportDefinition(definition), scope, {
+    [watermark.softDeleteColumn]: null,
+  });
+  return {
+    ...spec,
+    having: andHaving(spec.having, groupHaving),
```
willgriffin, line 352: The affected-group recompute filters by `HAVING <groupingColumn> = <value>` built from each changed-group row. When a grouping column is a **time bucket**, `groupHaving` keys off the bucketed output alias (e.g. `issued_month`) and value. Referencing a SELECT alias in `HAVING` and comparing a `date_trunc`/`strftime` expression to a stored value is dialect-sensitive (fine for sqlite `strftime` strings; postgres `date_trunc` yields a timestamp compared against a string).

`refresh-integration.test.ts` only covers plain-column grouping (`customerId`) — **incremental + time-bucket grouping is untested**. Worth a regression test for the bucketed incremental path before relying on it, since the time-bucket family is a headline feature of this package.

packages/reports/src/report.ts
```diff
@@ -0,0 +1,117 @@
...
+    const fields = await ObjectRegistry.getAllFields(reportClass);
+    const tenantField = [...fields.entries()].find(
+      ([fieldName, field]) =>
+        fieldName === 'tenantId' || field?._meta?.__tenancy?.isTenantIdField,
+    );
+    const tenantColumn = tenantField ? 'tenant_id' : null;
+    const tenantId = getTenantId() ?? null;
+    const result = tenantColumn
+      ? await this.db.query(
+          `SELECT MAX(refreshed_at) AS refreshed_at FROM ${tableName} WHERE ${tenantColumn} ${tenantId ? '= ?' : 'IS NULL'}`,
```
willgriffin: Two notes on `refreshIfStale`:
1. It hardcodes the `tenant_id` and `refreshed_at` column names, whereas `refresh.ts` resolves the tenant column from `tenantScopedConfig.field`. A report whose tenant field maps to a non-`tenant_id` column would mismatch here.
2. When a `ttl` is configured, every `list()`/`get()` pays an extra `MAX(refreshed_at)` round-trip, and a stale read triggers a **synchronous full refresh inside the read path**. That's a defensible design, but worth a doc note that TTL-on-read adds latency (and a write) to reads.

packages/reports/src/scheduler.ts
```diff
@@ -0,0 +1,604 @@
...
+  private startPolling(): void {
+    const poll = async () => {
+      if (!this.running) return;
+      try {
+        await this.poll();
+      } catch (error) {
+        this.emit('runner:error', error as Error);
+      }
+      if (this.running) {
+        this.pollTimer = setTimeout(poll, this.config.pollInterval);
```
willgriffin, line 474: `pollTimer` is created with `setTimeout` but never `unref()`d, so a running `ReportScheduleRunner` keeps the Node event loop (and process) alive until `stop()` is explicitly called. If the sibling `ScheduleRunner` in smrt-jobs has the same pattern, ignore; otherwise `this.pollTimer = setTimeout(...).unref()` avoids holding the process open.

packages/reports/vitest.config.ts
```diff
@@ -0,0 +1,19 @@
+import { defineConfig } from 'vitest/config';
+import { smrtVitestPlugin } from '../vitest/src/index.ts';
+
+export default defineConfig({
+  plugins: [smrtVitestPlugin({ verbose: true })],
+  test: {
+    globals: true,
+    environment: 'node',
+    include: ['src/**/*.test.ts'],
```
willgriffin: Per `.claude/rules/testing.md`: integration tests use `*.spec.ts` and should avoid mocking the database ("Never mock database operations; use real in-memory SQLite"). Here `refresh.test.ts` mocks the entire `DatabaseInterface` with `vi.fn()` to assert emitted SQL (testing generated output rather than the generator), and the real-DB `refresh-integration.test.ts` is named `.test.ts` rather than `.spec.ts`. This `include` glob also only matches `*.test.ts`, so any `*.spec.ts` integration test added later silently won't run. The integration coverage itself is excellent — this is just convention/naming alignment.

packages/reports/src/refresh.ts
willgriffin: Fixed in 53310fe73. Lock acquisition now uses a conditional INSERT ... ON CONFLICT DO UPDATE with the live-lock predicate on the write itself, then verifies the persisted owner_id before proceeding so concurrent losers skip refresh.

packages/reports/src/refresh.ts
willgriffin: Fixed in 53310fe73. Reports no longer read adapter internals; adapter type now comes from explicit options.adapterType or public db.url inference, and callers without an inferable URL must pass adapterType.

packages/reports/src/refresh.ts
willgriffin: Fixed in 53310fe73. Incremental refresh now asserts _smrt_report_watermarks even when trackRuns and lock are both disabled, and the SQLite integration suite includes the friendly missing-table regression.

packages/reports/src/refresh.ts
willgriffin: Fixed in 53310fe73. Added real SQLite incremental coverage for a monthly time-bucket report, including changed-group recompute against the bucketed issued_month grouping.

packages/reports/src/report.ts
willgriffin: Fixed in 53310fe73. refreshIfStale now resolves the report table, refreshedAt column, and tenant column from registry metadata instead of hardcoded column names, and ReportRefreshConfig.ttl documents the read-time query plus synchronous refresh cost.

packages/reports/src/scheduler.ts
willgriffin: Fixed in 53310fe73. ReportScheduleRunner now unrefs the poll timer when the runtime supports unref(), so the scheduler loop does not keep Node alive by itself.

packages/reports/vitest.config.ts
willgriffin: Fixed in 53310fe73. The SQLite integration suite is now *.spec.ts, Vitest includes specs, and the mocked refresh DB test was removed in favor of the real SQLite regressions.

packages/core/src/scanner/manifest-generator.ts
```diff
@@ -206,13 +208,23 @@ export class ManifestGenerator {
       }
     }
 
+    // Report cache rows are safe to scope by tenant even when a report is
+    // global: optional mode keeps tenant-less rows readable outside a tenant
+    // context and gives tenant-scoped reports the tenant_id column their raw
+    // aggregate refresh path must write.
+    this.normalizeReportTenantScope(manifest);
```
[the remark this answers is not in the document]
willgriffin: Fixed in 209480fdc. The Vite plugin now runs report tenant-scope normalization before tenant field injection and report object normalization before schema generation, matching generateManifest(); added a core manifest-generator regression for the manual pass order.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,1016 @@
...
+  return { watermarkColumn, softDeleteColumn };
+}
+
+function compileChangedGroupSpec(
+  definition: ReportDefinition,
+  scope: RefreshScope,
+  watermark: WatermarkConfig,
+  watermarkBefore: string,
+): AggregateSpec {
+  const spec = compileReportDefinition(definition);
```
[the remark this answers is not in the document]
willgriffin: Fixed in 209480fdc. Incremental refresh now ignores report.where while detecting changed groups, unions those groups with existing materialized groups when there is a current change or explicit change signal, and passes changedRows through scheduler/jobs for hard-delete paths. Added SQLite regressions for grouping-key moves, hard deletes with changedRows, and rows leaving a report filter.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,1016 @@
...
+async function rebuildReport(
+  db: DatabaseInterface,
+  definition: ReportDefinition,
+  reportTable: string,
+  scope: RefreshScope,
+  adapterType: SqlAdapterType,
+  runId: string | undefined,
+): Promise<RefreshWorkResult> {
+  const watermarkColumn =
+    definition.refresh?.watermarkColumn &&
```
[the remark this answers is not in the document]
willgriffin: Fixed in 209480fdc. The no-watermark seed rebuild now receives the resolved incremental watermark config, so default updatedAt/deletedAt are applied and _smrt_report_watermarks is written. Added a SQLite regression that verifies the first run seeds updated_at and the next run stays incremental.

packages/reports/src/refresh.ts
```diff
@@ -0,0 +1,1016 @@
...
+      `Report table '${reportTable}' does not exist for ${reportCtor.name}. ` +
+        `Run smrt db:migrate before refreshing reports.`,
+    );
+  }
+
+  if (options.trackRuns !== false || options.lock !== false) {
+    await assertReportTablesReady(options.db);
+  } else if (requestedMode === 'incremental') {
+    await assertReportTablesReady(options.db, [REPORT_WATERMARKS_TABLE]);
+  }
```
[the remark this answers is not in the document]
willgriffin: Fixed in 209480fdc. refreshReportOnce now derives required runtime tables from the selected mode and enabled features: runs only when tracking, watermarks for incremental, and locks only when locking. Added a lock:false regression with the locks table dropped.

Issue #1766, opened 2026-07-02
Epic: AdminShell — standardized WASD admin shell (App / Tenant / Focus / System)


willgriffin:
> New epic for the standardized admin app shell. Builds on the workspace primitives upstreamed by the (closed) #1226 lineage — `WorkspaceShell` / `RoleShell` / `ToolsDock` — and gives SMRT apps one spatial contract for the outer chrome. Informally: **the WASD shell**.

## Problem Statement

Every SMRT admin app (anytown, ergot, vesisos, the templates, the playground) re-invents its outer chrome. #1226 standardized inner workspace primitives, but apps still hand-roll headers, footers, and sidebars; keyboard ergonomics are ad hoc or absent; and there is no shared answer to "where does this kind of control live?"

For users that means relearning navigation with every SMRT app, digging through menus for tools that relate to what they're looking at, and having no glanceable view of system health (jobs, dispatch, environment) unless an app built one. For developers it means rebuilding the same shell plumbing per app, with no standard way for a page to contribute contextual tools and no ready-made user settings for any of it.

## Solution

One standardized admin shell where **each screen edge owns one scope of context**, ordered by altitude, and the **W/A/S/D keys toggle the edge they physically point at**:

- **Top = App scope (`W`)** — looks like a normal header until it drops down into an app panel: app/tenant switcher, environment, global search, docs/manual.
- **Left = Tenant scope (`A`)** — main navigation within the current tenant; slides between an icon rail and a labeled nav.
- **Right = Focus scope (`D`)** — a tool dock for the active page/record in the body: inspector, actions, AI operations, activity.
- **Bottom = System scope (`S`)** — looks like a normal status footer (environment, jobs, dispatch chips) until it slides up into a Stripe-workbench-style system panel.

Everything reads as a completely normal admin app until summoned. Motion is quick and light (transform-based, Stripe workbench feel). Push vs overlay is a per-app-per-panel design choice. Optional corner components let the header/footer visually fuse with the rails into full-height columns — or be omitted so header/footer content flows full-width. Users get shell settings (hotkeys off/remap, panel defaults) with bundled UI, persisted to localStorage by default with a tiny adapter seam for server persistence. Background processes are scope-bound **shell activities**: they surface as toasts on lifecycle events and minimize into the edge that owns them, which badges and hosts their progress.

## User Stories

1. As an admin user, I want the app to look like a normal admin app (header, footer, sidebars) until I summon more, so that the shell never gets in the way of the work in the body.
[...]

Issue #1779, opened 2026-07-02
admin-shell: shell activities — scope-bound background processes, edge badges, watch components


willgriffin:
## Parent

#1766

## What to build

Scope-bound **shell activities**: a client-side registry in the shell for background processes (a video encode, an import, a report rebuild). Each activity carries an owning scope — App / Tenant / Focus / System, with Focus activities holding a **subject** (the record/route they belong to) — plus a **kind** (e.g. `video-encode`, `import`, `report-rebuild`), lifecycle (started / progress / completed / failed / canceled), and optional determinate progress.

The owning scope's edge is the activity's **home**: it badges running work and hosts the scoped task list, while the system drawer additionally hosts the global list. When an app omits an activity's home edge, the activity re-homes to System.

Feeding follows the same seam discipline as the system-scope panels: a client API for app code plus a documented contract for mirroring server-side jobs (e.g. `_smrt_` jobs, video encodes) into activities — no new inter-package dependency.

Watch components ship with it: a per-edge **ActivityBadge** (running count, pulse, progress when determinate), **ActivityList** / **ActivityItem** (status, progress, cancel-action seam; embeddable in any pane with predicate filtering and hide-when-empty), a subject-scoped **activity dock tool** for the Focus edge, and a global activity panel for the system drawer.

Worked use case (ergot): a video-encode dock tool embeds an ActivityList filtered to running `video-encode` activities — hidden while nothing runs — so jobs started from that pane stay visible in that pane; the same activities also appear in the global system-drawer list.

## Acceptance criteria

- [ ] Registry tracks activities with scope (+ subject for Focus), kind, lifecycle transitions, and determinate/indeterminate progress; home-edge resolution falls back to System when the owning edge is omitted
- [ ] App feed seam: client API plus a documented contract for mirroring server-side jobs into activities; no new inter-package dependency
- [ ] ActivityBadge renders per edge (running count + pulse; progress ring when determinate) and updates live
- [ ] ActivityList/ActivityItem show status and progress with a cancel-action seam, accept a predicate filter (kind/scope/subject/status), support hide-when-empty, and embed inside any pane including app-authored dock tools; a global activity panel is available for the system drawer
- [ ] A subject-scoped activity dock tool registers through the focus-dock API and shows the current subject's activities
[...]

Issue #1777, opened 2026-07-02
admin-shell: adoption — template-sveltekit default shell + WorkspaceShell migration guide


willgriffin:
## Parent

#1766

## What to build

Make the WASD shell the default SMRT admin experience: **template-sveltekit** scaffolds with AdminShell chrome — manifest-driven tenant nav wired, WASD hotkeys on, and a settings page including the ShellSettingsPanel — plus a **migration guide** for existing WorkspaceShell/RoleShell consumers showing the additive path (no breaking changes), and a playground reference demo covering all four scopes linked from the docs.

## Acceptance criteria

- [ ] template-sveltekit scaffolds with AdminShell chrome: manifest-driven tenant nav, WASD hotkeys enabled, and a settings page with ShellSettingsPanel
- [ ] Migration guide documents moving from WorkspaceShell/RoleShell composition to AdminShell additively; existing consumers keep working unchanged
- [ ] Playground reference demo covers all four scopes and is linked from the docs
- [ ] Template build, typecheck, and tests are green; knowledge-freshness checks pass

## Blocked by

- #1767
- #1768
- #1772
- #1773

PR #1786, opened 2026-07-02
[codex] AdminShell standardized workspace release

willgriffin:
## Summary

- Add the canonical `@happyvertical/smrt-svelte/workspace` AdminShell release: four-edge shell state/config, physical WASD hotkeys, settings adapter/UI, focus tools, activities/toasts, system/app/tenant panels, and manifest tenant nav.
- Document the AdminShell layout contract, migration path, and release decisions so the new pattern is not constrained by the deprecated shell.
- Update the playground and SvelteKit template adoption so new apps start on the standard AdminShell.
- Add a separate mechanical Biome baseline phase required by the repo pre-push hook.

Refs #1766

## Validation

- `pnpm --filter @happyvertical/smrt-svelte typecheck`
- `pnpm --filter @happyvertical/smrt-svelte test`
- `pnpm --filter @happyvertical/smrt-svelte build`
- `pnpm --filter @happyvertical/smrt-svelte verify:pack`
- `pnpm --filter @happyvertical/smrt-template-sveltekit test`
- `pnpm --filter @happyvertical/smrt-core typecheck`
- `pnpm --filter @happyvertical/smrt-content test`
- `pnpm run check:svelte-tokens`
- `pnpm run check:hardcoded-strings`
- `pnpm run check:raw-primitives`
- `pnpm run check:color-literals`
- `pnpm run check:z-index`
- `pnpm run check:spacing`
- `pnpm run check:radius`
- `pnpm run check:elevation`
- `pnpm run check:typography`
- `pnpm run check:standards`
- `pnpm run check:package-dag`
- `pnpm run check:no-smrt-peers`
- `pnpm run knowledge:check -- --changed --strict --format markdown`
- `pnpm run build`
- `pnpm test`
- `pnpm run typecheck` (standalone rerun; first parallel run raced generated build output)
- `npx biome ci --diagnostic-level=error .`
- `git push -u origin codex/1766-admin-shell-release` (pre-push hook passed)


packages/smrt-svelte/src/components/workspace/admin-shell/hotkeys.ts
```diff
@@ -0,0 +1,81 @@
...
+export function shouldIgnoreShellHotkey(
+  event: KeyboardEvent,
+  settings: ShellSettingsDelta = {},
+): boolean {
+  if (settings.hotkeysEnabled === false) return true;
+  if (event.isComposing) return true;
+  if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
+    return !isShortcutsEvent(event);
+  }
+  return isEditableTarget(event.target);
```
willgriffin: **`?` shortcut bypasses the editable-target guard.** When any modifier (including Shift) is held, this returns `!isShortcutsEvent(event)` at line 39 and never reaches the `isEditableTarget` check at line 41. Since `?` is Shift+/, pressing it while focused in an `<input>`/`<textarea>`/`contenteditable` returns `false` (not ignored) → `AdminShell.handleKeydown` opens the overlay **and** `event.preventDefault()`s the character, so the user can't type `?` into any field in the shell.

This is the exact case story 6 calls out ("single-letter hotkeys suppressed while an editable control has focus"). Suggest checking `isEditableTarget(event.target)` before the shortcuts short-circuit, e.g. return `true` early when the target is editable regardless of modifiers.

packages/smrt-svelte/src/components/workspace/admin-shell/ActivityToasts.svelte
```diff
@@ -0,0 +1,127 @@
...
+}
+
+let { max = 4 }: Props = $props();
+const { t } = useI18n();
+const shell = useAdminShell();
+let toasts = $state<Toast[]>([]);
+let unwatch: (() => void) | null = null;
+
+function pushToast(event: ShellActivityEvent): void {
+  if (event.type === 'remove') return;
```
willgriffin: **Toast fires on every non-`remove` event, including progress-only updates.** `updateActivity({ progress })` keeps `status` unchanged, so `upsertActivity` emits a `type: 'upsert'` event, and `pushToast` only skips `remove` — so a running activity that reports progress (the canonical video-encode from #1779/#1780) spawns a fresh toast on **every progress tick**. `updatedAt` changes each call, so the `(toast.id)` key churns and the stack re-renders continuously.

ADR 0002 (line 37) says the toast component should "decide which events notify." Suggest filtering to status transitions (and/or an app-configurable event set), e.g. only toast when `event.type === 'transition'` or on an explicit started/completed/failed allowlist.

packages/smrt-svelte/src/components/workspace/admin-shell/AdminShell.svelte
```diff
@@ -0,0 +1,643 @@
...
+  }
+</script>
+
+{#snippet edgeToggle(edge: PanelEdge)}
+  <Button
+    variant="ghost"
+    size="sm"
+    class="smrt-admin-shell__edge-toggle"
+    aria-expanded={edgeExpanded(edge)}
+    aria-controls={`smrt-admin-shell-${edge}-panel`}
```
willgriffin, line 218: **`aria-controls` dangles for the left/right edges.** `edgeToggle` emits `aria-controls="smrt-admin-shell-${edge}-panel"` for all four edges, but only `top` (id at line 237) and `bottom` (id at line 327) render an element with that id. The tenant (`left`) and focus (`right`) rails render inline with no `smrt-admin-shell-left-panel` / `-right-panel` element, so the toggle references a non-existent id and `aria-expanded` points at nothing a screen reader can resolve. Either give the rail content the matching id, or only emit `aria-controls` for edges that render a drawer section.

packages/smrt-svelte/src/components/workspace/admin-shell/AdminShell.svelte
```diff
@@ -0,0 +1,643 @@
...
+        </section>
+      {/if}
+    </footer>
+  {/if}
+
+  {#if shortcutsOpen}
+    <div
+      class="smrt-admin-shell__shortcuts"
+      role="dialog"
+      aria-modal="true"
```
willgriffin, line 393: **`aria-modal="true"` with no focus management.** The shortcuts dialog never moves focus into itself, doesn't trap Tab, and doesn't restore focus to the opener on close — so `aria-modal="true"` tells assistive tech the background is inert when it isn't, and keyboard focus stays on the page behind. ADR 0002 (line 7) lists focus trapping as expected post-mount behavior. Suggest autofocusing the panel on open, containing Tab, and restoring focus to the trigger on close (Escape handling at line 116 already covers dismissal).

packages/smrt-svelte/src/components/workspace/admin-shell/HotkeyInput.svelte
```diff
@@ -0,0 +1,69 @@
...
+
+interface Props {
+  value?: ShellHotkeyBinding | null;
+  disabled?: boolean;
+  conflictsWith?: string | null;
+}
+
+let { value = null, disabled = false, conflictsWith = null }: Props = $props();
+const { t } = useI18n();
+const dispatch = createEventDispatcher<{
```
willgriffin: **Deprecated Svelte 5 event API.** `createEventDispatcher` (here) plus the `on:capture` directive at `ShellSettingsPanel.svelte:84` are both deprecated in Svelte 5 runes mode. It works today (smrt-ui `Button` forwards `onkeydown` via `{...rest}`, and legacy events still fire), so this is convention not breakage — but it's the only Svelte-4 event pattern in an otherwise runes-native family. Suggest a callback prop, e.g. `oncapture?: (binding: ShellHotkeyBinding | null) => void`, called directly instead of `dispatch('capture', …)`.

willgriffin (commented):
## Review — Step 1 (Phase 0 decisions + Phase 1 AdminShell core)

Scope: the AdminShell core surface — `state.svelte.ts`, `hotkeys.ts`, `settings.ts`, `types.ts`, the shell + activity components, ADR 0002, and CONTEXT.md.

**Reconciliation note:** I first reviewed the phase-1 commit (`372abe5d`) in isolation and found z-index/i18n ratchet failures — then re-verified against the PR head. **Later phases already fixed those**, so I've dropped them: `check:z-index`, `check:hardcoded-strings`, `check:typography`, and `check:raw-primitives` all pass clean at head. The five findings below are the ones that **survive at the PR head** (`0002853e`).

Solid work overall — the state machine, sparse-delta settings, and physical-`code` keymap are clean and well-tested. Findings are inline; most severe first:

1. **(correctness)** `hotkeys.ts` — `?` opens the shortcuts overlay even while typing in an input/textarea.
2. **(correctness)** `ActivityToasts.svelte` — a toast fires on *every* activity update, including progress ticks (storm on a running encode).
3. **(a11y)** `AdminShell.svelte:174` — `aria-controls` dangles for the left/right edge toggles.
4. **(a11y)** `AdminShell.svelte:347` — shortcuts dialog is `aria-modal` with no focus trap/restoration.
5. **(svelte5 convention)** `HotkeyInput.svelte` — deprecated `createEventDispatcher` + `on:capture`.

### Design decision to confirm (not a defect)
ADR 0002 (lines 15–17) makes top/bottom drawer **mutual exclusion opt-in** via `exclusiveGroup`, and the default config sets no group — so by default opening the System drawer does **not** close the App drawer. This deliberately supersedes the epic's locked decision ("the top and bottom drawers are mutually exclusive (opening one closes the other)"). The code/tests/ADR are internally consistent; I'm only flagging that it reverses a locked epic decision, so confirm the opt-in default is what you want (or set `{ top: { exclusiveGroup: 'vertical' }, bottom: { exclusiveGroup: 'vertical' } }` in the template/playground defaults).

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
```diff
@@ -0,0 +1,312 @@
...
+  private closeExclusivePeers(edge: PanelEdge): void {
+    const group = this.config.panels[edge].exclusiveGroup;
+    if (!group) return;
+    for (const peer of PANEL_EDGES) {
+      if (
+        peer !== edge &&
+        this.config.panels[peer].exclusiveGroup === group &&
+        this.panels[peer] === 'expanded'
+      ) {
+        this.panels[peer] = 'collapsed';
```
willgriffin, line 271: **Exclusivity close isn't persisted, so it's undone on reload.** `closeExclusivePeers` mutates `this.panels[peer]` directly, but only the *toggled* edge is written to `this.settings.panels` (in `setPanel`, line 116-118) — the collapsed peer isn't. Scenario with `{ top: { exclusiveGroup: 'vertical' }, bottom: { exclusiveGroup: 'vertical' } }`: expand top (persists `panels.top='expanded'`), then expand bottom → this line collapses top in memory but `settings.panels.top` stays `'expanded'`. After reload, `resolveInitialPanelState` restores **both** to `'expanded'`, violating the exclusivity the group was configured for. Suggest routing the peer collapse through the same settings merge (or have `setPanel` persist affected peers) so the persisted state matches the exclusivity rule.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
```diff
@@ -0,0 +1,312 @@
...
+          : determinate.reduce((total, progress) => total + progress, 0) /
+            determinate.length,
+    };
+  }
+
+  homeEdgeForScope(scope: ShellScope): PanelEdge {
+    const preferred = SCOPE_EDGES[scope];
+    if (this.panels[preferred] !== 'hidden') return preferred;
+    if (this.panels.bottom !== 'hidden') return 'bottom';
+    return preferred;
```
willgriffin: **`homeEdgeForScope` can return a hidden edge → activity homes to nowhere.** If the scope's preferred edge is hidden and `bottom` is *also* hidden (e.g. an app configures `{ right: false, bottom: false }`), this falls through to `return preferred` — a hidden edge. The activity then carries `edge: 'right'`, `activityBadge('right')` renders nothing (edge hidden), and it's absent from the system list too, so a running process is invisible. Suggest falling back to the first non-hidden edge (or an explicit visible default) instead of the hidden preferred edge. The re-home test only covers `{ right: false }` with `bottom` visible, so this path is untested.

packages/smrt-svelte/src/components/workspace/admin-shell/SystemScopePanel.svelte
```diff
@@ -0,0 +1,112 @@
...
+  {#each panels as panel (panel.id)}
+    <section class="smrt-system-scope-panel__group" aria-label={panel.label}>
+      <header>
+        <h2>{panel.label}</h2>
+        <span>{panel.items.length}</span>
+      </header>
+      {#if panel.items.length > 0}
+        <div class="smrt-system-scope-panel__items">
+          {#each panel.items as item (item.id)}
+            <a href={item.href ?? '#'} class:disabled={!item.href}>
```
willgriffin: **"Disabled" system links are still keyboard-activable.** `class:disabled={!item.href}` only applies `pointer-events: none` (mouse), and `href='#'` keeps the `<a>` in the tab order and Enter-activable — so a keyboard user can focus a status item with no destination and navigate to `#` (jumps to top / dirties the URL). There's no `aria-disabled`, so assistive tech announces it as a normal link. Suggest rendering a non-interactive element (e.g. `<span>`/`<div>`) when `!item.href`, or adding `aria-disabled="true"` and preventing keyboard activation.

packages/smrt-svelte/src/components/workspace/admin-shell/ActivityItem.svelte
```diff
@@ -0,0 +1,80 @@
...
+    {/if}
+    {#if typeof activity.progress === 'number'}
+      <progress max="100" value={activity.progress}>
+        {Math.round(activity.progress)}%
+      </progress>
+    {/if}
+  </div>
+  <div class="smrt-activity-item__actions">
+    {#if activity.detailHref}
+      <Button variant="ghost" size="sm" href={activity.detailHref}>View</Button>
```
willgriffin: **Two user-facing strings bypass i18n.** `View` here and `Cancel` at line 31 are hardcoded, while every other shell string routes through `t(M[...])` (and `ShortcutsOverlay.svelte:49` has a bare `Esc`). `check:hardcoded-strings` doesn't flag them because its PROSE heuristic only matches ≥2-word text — single words slip through — so the ratchet passing is a false negative here, not a clean bill. For the standard shell that every SMRT app inherits, these should go through `strings.workspace` like the rest.

willgriffin (commented):
## Review — round 2 (whole PR at head)

Second pass covering the phase-2 content components (TenantNav, SystemScopePanel, AppScopePanel, SystemStatusChips, Activity*), the barrel/exports, the manifest tenant-nav, and the template/playground adoption — all verified at head (`0002853e`).

### Headline: ~147 files of unrelated formatting churn are bundled into this release
The phase-4 commit (`chore(format): satisfy pre-push biome baseline`, `c66a1a41`) reformats **~147 files that have nothing to do with AdminShell** — `packages/content` alone is **115 files / ~1,660 lines**, plus `core`, `facts`, `users`, `features`, `cli`, `secrets`, `scanner`, `profiles`, `products`, `config`. Sampling several (`content/.../submit/+server.ts`, `content/src/body-format.ts`, `core/*`) shows pure Biome churn — `"x"`→`'x'` quote normalization, line re-wrapping, blank-line removal, a `biome-ignore` reflow — no logic changes.

This is roughly half the PR's 9,162-line diff and I'd split it out: it (a) makes the shell review needle-in-haystack, (b) pollutes the `smrt-svelte` release changelog with content/core edits, (c) couples a revert of the shell to a revert of the formatting, and (d) suggests a Biome version/config skew vs `main` (these files were "clean" on main). Recommend landing the format baseline as its own maintenance PR (or dropping it here) and keeping #1786 to `smrt-svelte` + `template-sveltekit` + docs.

### Code findings (inline)
- `state.svelte.ts:267` — exclusivity close isn't persisted → both panels restore expanded after reload.
- `state.svelte.ts:255` — `homeEdgeForScope` can return a hidden edge, homing an activity to nowhere.
- `SystemScopePanel.svelte:29` — "disabled" links stay keyboard-activable.
- `ActivityItem.svelte:27` — two user-facing strings escape i18n (ratchet can't catch single words).

### Test coverage gaps (the risky logic is under-tested)
The suite covers happy paths but not the edges where the bugs live: (1) no test that `?` is **suppressed** while an input is focused — `admin-shell-hotkeys.test.ts` fires `?` with no focused target, so it passes while the real bug ships; (2) no test that an exclusivity-closed peer stays collapsed **after re-hydrate**; (3) no test that a progress-only `updateActivity` does/doesn't spawn a toast; (4) no test for `homeEdgeForScope` when both the scope edge and `bottom` are hidden. Adding these four would have caught findings 1, 2, and the toast storm from round 1.

### Clean / verified (no action)
Manifest tenant-nav (`manifest-nav.ts`, pre-existing) is a solid pure helper with STI-parent permission expansion; `package.json` exports/`files` are unchanged (no new subpath, so no `smrt-vitest` alias gap); AdminShell defaults `storageKey`, so template persistence works; keydown listener + hydrate are correctly `onMount`-guarded for SSR; `ShellDockTool` register/unregister lifecycle is correct.

packages/smrt-svelte/src/components/workspace/admin-shell/hotkeys.ts
willgriffin: Fixed in a677d47a8 — the editable-target check now runs before the modifier branch, so `?` typed in an input/textarea is suppressed. Added a regression test.

packages/smrt-svelte/src/components/workspace/admin-shell/ActivityToasts.svelte
willgriffin: Fixed in a677d47a8 — toasts fire only on lifecycle changes (new activity or status transition); progress-only ticks are filtered. Added a `notify` predicate prop + regression test.

packages/smrt-svelte/src/components/workspace/admin-shell/AdminShell.svelte
willgriffin: Fixed in a677d47a8 — aria-controls now targets the persistent edge containers (header/aside/aside/footer), which always exist, so left/right no longer dangle.

packages/smrt-svelte/src/components/workspace/admin-shell/AdminShell.svelte
willgriffin: Fixed in a677d47a8 — added a focus trap + focus restoration to the shortcuts dialog (use:focusTrap).

packages/smrt-svelte/src/components/workspace/admin-shell/HotkeyInput.svelte
willgriffin: Fixed in a677d47a8 — HotkeyInput and ShortcutsOverlay now use callback props (oncapture / onclose) instead of createEventDispatcher + on:.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
willgriffin: Fixed in a677d47a8 — closeExclusivePeers now persists the collapsed peer into settings; added a re-resolve regression test.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
willgriffin: Fixed in a677d47a8 — falls back to the first visible edge when the scope edge and System are both hidden; added a both-hidden regression test.

packages/smrt-svelte/src/components/workspace/admin-shell/SystemScopePanel.svelte
willgriffin: Fixed in a677d47a8 — hrefless status items now render as non-interactive elements with aria-disabled instead of keyboard-activable # links.

packages/smrt-svelte/src/components/workspace/admin-shell/ActivityItem.svelte
willgriffin: Fixed in a677d47a8 — View/Cancel now routed through i18n (ui.activity_item.view / .cancel).

packages/smrt-svelte/src/components/workspace/admin-shell/hotkeys.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export function shouldIgnoreShellHotkey(
+  event: KeyboardEvent,
+  settings: ShellSettingsDelta = {},
+): boolean {
+  if (settings.hotkeysEnabled === false) return true;
+  if (event.isComposing) return true;
+  if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
+    return !isShortcutsEvent(event);
+  }
```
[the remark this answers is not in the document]
willgriffin: Deferred by design: ADR 0002 scopes this release to unmodified physical keys for panel toggles ("the binding type should remain extensible enough to add modifier chords later"). The guard intentionally ignores modifier combos for now; `setHotkey` stores a bare `code`. Modifier-chord support is future work, not a regression here.

packages/smrt-svelte/src/components/workspace/admin-shell/ShellSettingsPanel.svelte
```diff
@@ -0,0 +1,146 @@
...
+              {shell.panels[edge] === 'expanded'
+                ? t(M['ui.shell_settings_panel.collapse'])
+                : t(M['ui.shell_settings_panel.expand'])}
+            </Button>
+            <HotkeyInput
+              value={bindingFor(edge)}
+              disabled={!hotkeysEnabled}
+              conflictsWith={conflictFor(edge)}
+              on:capture={(event) =>
+                shell.setHotkey(edge, event.detail?.code ?? null)}
```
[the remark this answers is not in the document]
willgriffin: Same scope call as above — the current release stores only `code` (`setHotkey(edge, code)`) because modifier chords are deferred per ADR 0002. HotkeyInput keeps the full binding shape so the storage model can extend later without a rewrite. Outdated after the rebase; resolving.

packages/template-sveltekit/template/src/routes/+page.svelte
```diff
@@ -1,8 +1,47 @@
...
+  const navItems: ShellNavItem[] = [
+    { href: '/', label: 'Items', icon: 'I' },
+    { href: '/api/items', label: 'API', icon: 'A' },
+  ];
+
+  const statusChips: ShellStatusChip[] = [
+    { id: 'env', label: 'Local', tone: 'info' },
+    { id: 'items', label: 'Items', value: items.length },
+    { id: 'connection', label: error ? 'API error' : 'Ready', tone: error ? 'error' : 'success' },
+  ];
```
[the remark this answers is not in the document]
willgriffin: Moot here: the template-sveltekit adoption was dropped from this PR during the rebase onto main (main's server-load template #1781 conflicts architecturally). AdminShell template adoption is deferred to slice #1777, where it'll be built against the current template. Resolving.

packages/smrt-svelte/src/components/workspace/admin-shell/HotkeyInput.svelte
```diff
@@ -0,0 +1,69 @@
...
+  }
+  if (!event.code) return;
+  dispatch('capture', {
+    code: event.code,
+    altKey: event.altKey || undefined,
+    ctrlKey: event.ctrlKey || undefined,
+    metaKey: event.metaKey || undefined,
+    shiftKey: event.shiftKey || undefined,
+  });
+}
```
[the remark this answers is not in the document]
willgriffin: Fixed in 2679d49fa — `handleKeydown` now returns early on Tab / Shift+Tab so focus can leave the capture control instead of being trapped by the unconditional preventDefault. (Tab therefore can't be bound as a hotkey, an acceptable trade for keyboard escape.)

packages/smrt-svelte/src/components/workspace/admin-shell/SystemScopePanel.svelte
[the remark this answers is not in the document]
willgriffin: Fixed in 9e2820e6c — hrefless status items now render as a non-interactive `<span>` with `aria-disabled` (via `<svelte:element>`) instead of a focusable `href="#"` link, so keyboard users can't activate a no-destination item.

packages/smrt-svelte/src/components/workspace/admin-shell/ActivityItem.svelte
```diff
@@ -0,0 +1,80 @@
...
+    {/if}
+    {#if activity.cancel && (activity.status === 'queued' || activity.status === 'running')}
+      <Button variant="ghost" size="sm" onclick={() => activity.cancel?.()}>
+        Cancel
+      </Button>
+    {/if}
```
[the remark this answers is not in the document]
willgriffin: Fixed in 9e2820e6c — View/Cancel now come from the workspace catalog (`ui.activity_item.view` / `.cancel`) via `useI18n().t`.

packages/smrt-svelte/src/components/workspace/admin-shell/ActivityList.svelte
```diff
@@ -0,0 +1,40 @@
...
+  filter?: ShellActivityFilter;
+  hideWhenEmpty?: boolean;
+  emptyLabel?: string;
+}
+
+let {
+  filter = {},
+  hideWhenEmpty = false,
+  emptyLabel = 'No activities',
+}: Props = $props();
```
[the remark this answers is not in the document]
willgriffin: Fixed in 2679d49fa — the default empty-state label is now routed through the workspace catalog (`ui.activity_list.empty`); callers can still override via the `emptyLabel` prop.

PR #1833, opened 2026-07-04
feat(users): add operation permission guards

willgriffin:
## Summary
- add reusable `assertOperationPermission` / `checkOperationPermission` / `hasOperationPermission` guards for hand-written write surfaces
- derive operation permission slugs from the users permission catalog and fail closed for unknown catalog entries
- preserve system-context and super-admin bypass behavior, with explicit suppression options where needed
- add catalog-aware default role-permission seeding and opt-in `seedSystemRoles({ seedPermissions: true })`
- document guard/seeding usage and add focused users tests
- include two mechanical Biome fixes required by the root pre-push format gate after rebasing on #1787

Closes #1829

## Validation
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm exec turbo build --force --env-mode=loose`
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm --filter @happyvertical/smrt-users test -- operation-permission`
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm --filter @happyvertical/smrt-users typecheck`
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm exec turbo typecheck --env-mode=loose`
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm exec turbo test --env-mode=loose`
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm exec turbo lint --env-mode=loose`
- `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0 PNPM_CONFIG_DANGEROUSLY_ALLOW_ALL_BUILDS=true pnpm exec turbo format-check --env-mode=loose`
- `npx biome ci --diagnostic-level=error .`
- `node_modules/.bin/tsx scripts/check-knowledge.ts --strict --format markdown`
- `git diff --check`
- SMRT review pass via `smrt_review` returned only deterministic routing warnings and no concrete findings


packages/users/src/collections/RolePermissionCollection.ts
```diff
@@ -5,6 +5,106 @@
...
+  unmatchedPatterns: Record<string, string[]>;
+}
+
+export const DEFAULT_ROLE_PERMISSION_PATTERNS: Record<
+  DefaultRoleSlug,
+  readonly string[]
+> = {
+  [DEFAULT_ROLE_SLUGS.OWNER]: ['*'],
+  [DEFAULT_ROLE_SLUGS.ADMIN]: ['*'],
+  [DEFAULT_ROLE_SLUGS.MEMBER]: ['*.read', '*.list', '*.get', '*.create'],
```
willgriffin, line 52: **Privilege escalation: this `member` default grants write on the RBAC/authority collections.**

Traced Codex's P1 end-to-end on this branch. `*.create` matches every `<collection>.create` slug in the catalog — including the users authority models. Those are API+MCP read-only (#1400) but still ship `cli: true`:

```
Role / Permission / Membership / MembershipOverride / Group
  api: { include: ['list','get'] }, mcp: { include: ['list','get'] }, cli: true
```

`PermissionCatalogService.getManifestPermissionDefinitions()` derives the create/update/delete slug from the **union** of api|cli|mcp exposure:

```ts
const exposed =
  isOperationEnabled(objectConfig.api, action) ||   // false — include lacks 'create'
  isOperationEnabled(objectConfig.cli, action) ||   // true  — cli === true, no include filter
  isOperationEnabled(objectConfig.mcp, action);     // false
```

so `roles.create`, `permissions.create`, `memberships.create`, `membership_overrides.create`, `groups.create` all land in the catalog and get granted to `member` on `seedRolePermissions()` / `seedSystemRoles({ seedPermissions: true })`.

Exploit — using the guard this PR exists to enable:

```ts
await assertOperationPermission({ collection: MembershipOverride, action: 'create', userId, tenantId });
```

A plain member passes it and can mint a grant-effect override for **any** permission → full self-escalation. `memberships.create` (self-assign a role) and `roles.create` are equivalent vectors. This re-opens on the hand-written surface exactly what #1400 closed on the generated surface.

Fix options (any one):
1. Exclude the users authority collections from the `member` default (drop their `.create`), or restrict `member` to `*.read` and make create opt-in per app.
2. Better long-term: teach the catalog to distinguish CLI-only writes from REST-exposed writes, so a `cli:true` / REST-read-only model doesn't emit a `.create` slug that a web guard will honor — otherwise every future `cli:true` authority model inherits this footgun.

Suggested regression test: seed defaults, then assert `member` does **not** resolve `membership_overrides.create` / `roles.create` / `memberships.create`.

packages/users/src/services/OperationPermissionService.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+async function readTenancyBypassContext(): Promise<RuntimeBypassContext> {
+  try {
+    const tenancy = await import('@happyvertical/smrt-tenancy');
+    return {
+      superAdminBypass: tenancy.isSuperAdminBypass(),
+      systemContext: tenancy.isSystemContext(),
+    };
+  } catch {
```
willgriffin: Endorsing Copilot's note here. This bare `catch {}` swallows *every* import failure and reports "no bypass," diverging from the sibling `SessionPermissionContext.ts` (`isModuleNotFoundError` at L130, used at L148/L173) which only treats `ERR_MODULE_NOT_FOUND` as "tenancy is optional" and rethrows everything else.

Direction-wise it fails **closed** — a genuine init/runtime error in `@happyvertical/smrt-tenancy` silently demotes a real system/super-admin context to "denied" — so it's not itself a vuln, but it converts a real load error into a silent, hard-to-diagnose authorization failure.

Suggest reusing the same guard: swallow module-not-found → `{ superAdminBypass: false, systemContext: false }`, and surface anything else (rethrow, or fold into a `resolution_error` deny so `checkOperationPermission` still returns a decision rather than throwing out of `hasOperationPermission`).

packages/users/src/collections/RolePermissionCollection.ts
```diff
@@ -88,4 +188,99 @@ export class RolePermissionCollection extends SmrtCollection<RolePermission> {
...
+      const sortedMatchedSlugs = Array.from(matchedSlugs).sort();
+      result.matched[roleSlug].push(...sortedMatchedSlugs);
+
+      const existingPermissionIds = new Set(
+        await this.getPermissionIds(role.id),
+      );
+      const targetPermissionIds = new Set<string>();
+
+      for (const slug of sortedMatchedSlugs) {
+        const permission = await permissions.findBySlug(slug);
```
willgriffin: Minor / optional — seeding is one-shot so the impact is low. This is an N+1: `findBySlug` per matched slug, and for `owner`/`admin` (`['*']`) that's one query per catalog permission, per role. Could preload once (e.g. `permissions.list({})` → `Map<slug, id>`) and do O(1) lookups while iterating. Fine to defer to a follow-up.

willgriffin (commented):
Reviewed the diff end-to-end and verified both bot findings against **this branch's** actual model configs + catalog derivation (my local checkout was on a stale branch, so I re-checked every claim against `origin/codex/issue-1829-write-authz`). The feature itself is clean, fail-closed by construction, and genuinely well-tested — nice work. One blocker before merge, plus two smaller notes, all inline.

**🔴 Must fix — privilege escalation in the default seeding matrix (agrees with Codex's P1, verified end-to-end).** The `member` → `*.create` default grants write on the RBAC/authority collections (`roles` / `permissions` / `memberships` / `membership_overrides` / `groups`), because those models are REST+MCP read-only but still `cli: true`, and the catalog derives `.create` from the **union** of api|cli|mcp exposure. Any app that adopts the documented default seeding and then uses this PR's own guard can be self-escalated by an ordinary member. Mechanism + exploit path + fix options + a suggested regression test are inline on the pattern definition.

**🟡 Endorse — Copilot's tenancy `catch` note.** Real robustness/consistency issue: it fails *closed* (so not itself a vuln) but silently masks genuine `@happyvertical/smrt-tenancy` load errors and diverges from the sibling `SessionPermissionContext.ts`. Inline.

**⚪ Minor — Copilot's N+1 note.** Real but low-impact (seed-time only). Inline.

Non-issue I checked so you don't have to: the `resolver.hasPermission(..., { membership })` call is correct — `PermissionResolutionOptions` already carries `membership` on the base branch's `PermissionResolver`, so there's nothing to change there.

Net: I'd hold merge on the `member` default only; the other two are optional/follow-up.

packages/users/src/services/OperationPermissionService.ts
```diff
@@ -0,0 +1,208 @@
...
+    return {
+      superAdminBypass: false,
+      systemContext: false,
+    };
+  }
+}
```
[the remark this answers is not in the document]
willgriffin: Addressed in a9f025e39. `readTenancyBypassContext()` now only treats missing `@happyvertical/smrt-tenancy` as optional, rethrows unexpected import/runtime failures, and `checkOperationPermission()` folds those into a `resolution_error` decision instead of silently denying. Validated with `pnpm --filter @happyvertical/smrt-users test -- operation-permission`, users typecheck, Biome CI on touched files, and knowledge freshness.

packages/users/src/collections/RolePermissionCollection.ts
```diff
@@ -88,4 +188,99 @@ export class RolePermissionCollection extends SmrtCollection<RolePermission> {
...
+
+        targetPermissionIds.add(permission.id);
+        if (existingPermissionIds.has(permission.id)) {
+          result.unchanged[roleSlug].push(slug);
+          continue;
+        }
+
+        await this.addPermission(role.id, permission.id);
+        result.added[roleSlug].push(slug);
+      }
```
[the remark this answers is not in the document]
willgriffin: Addressed in a9f025e39. `seedRolePermissions()` now preloads `permissions.list({})` into a slug-to-id map once, then uses O(1) lookups while iterating matched slugs instead of calling `findBySlug()` for each slug. Validated with `pnpm --filter @happyvertical/smrt-users test -- operation-permission`, users typecheck, Biome CI on touched files, and knowledge freshness.

packages/users/src/collections/RolePermissionCollection.ts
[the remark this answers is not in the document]
willgriffin: Addressed in a9f025e39. The member default still allows create for ordinary app resources, but now excludes users/RBAC authority and security resources from member `.create` grants. Added a regression proving owner/admin still receive those create slugs while member does not. Docs updated. Validated with `pnpm --filter @happyvertical/smrt-users test -- operation-permission`, users typecheck, Biome CI on touched files, and knowledge freshness.

willgriffin: Addressed in a9f025e39. I kept member create for ordinary app resources but denylisted the users/RBAC authority and security resources for default member seeding. The emitted catalog slugs for several of these are compact (`membershipoverrides.create`, `rolepermissions.create`, etc.), so the implementation blocks both compact and underscored resource forms. The regression seeds defaults with all users models registered and asserts owner/admin receive the authority create slugs while member does not. Docs were updated with the narrowed default. Validated with `pnpm --filter @happyvertical/smrt-users test -- operation-permission`, users typecheck, Biome CI on touched files, and knowledge freshness.

packages/users/src/services/OperationPermissionService.ts
willgriffin: Addressed in a9f025e39. The guard now mirrors the sibling behavior: missing `@happyvertical/smrt-tenancy` is the only swallowed import failure; unexpected load/runtime failures are surfaced and returned from `checkOperationPermission()` as `resolution_error`, so `hasOperationPermission()` remains boolean/fail-closed without masking the underlying error. Validated with `pnpm --filter @happyvertical/smrt-users test -- operation-permission`, users typecheck, Biome CI on touched files, and knowledge freshness.

packages/users/src/collections/RolePermissionCollection.ts
willgriffin: Addressed in a9f025e39. I preloaded permissions once into a `Map<slug, id>` before the role loop and removed the per-slug `findBySlug()` query. Validated with `pnpm --filter @happyvertical/smrt-users test -- operation-permission`, users typecheck, Biome CI on touched files, and knowledge freshness.

PR #1839, opened 2026-07-04
fix(core): align endpoint previews with API policies

willgriffin:
## Summary

Addresses the unresolved review feedback from #1838 for `previewApiEndpoints`:

- exclude sensitive fields from list filter parameter previews and generated list curl examples
- mirror generated REST writable policy for create/update body params (`id`, tenant fields, timestamps, `_` fields, readonly fields, and `api.writable` allowlists)
- generate UUID-shaped examples for relationship/id fields while letting declared numeric/boolean/date types win
- make `[in]` comma-separated query syntax discoverable in parameter descriptions
- switch the default curl host to `https://api.example.com`, add `baseUrl`, and shell-quote JSON bodies safely

## Validation

- `CI=true fnm exec --using 24.18.0 pnpm exec biome check packages/core/src/mcp-advisor/types.ts packages/core/src/mcp-advisor/tools/preview-api-endpoints.ts packages/core/src/mcp-advisor/tools/preview-api-endpoints.test.ts packages/core/src/__tests__/fixtures/advisor-test-classes.ts`
- `CI=true fnm exec --using 24.18.0 pnpm --filter @happyvertical/smrt-core exec vitest run src/mcp-advisor/tools/preview-api-endpoints.test.ts`
- `CI=true fnm exec --using 24.18.0 pnpm --filter @happyvertical/smrt-core exec vitest run src/mcp-advisor`
- `CI=true fnm exec --using 24.18.0 pnpm --filter @happyvertical/smrt-core typecheck`
- `CI=true fnm exec --using 24.18.0 pnpm run build`
- `CI=true fnm exec --using 24.18.0 pnpm run typecheck`
- `CI=true fnm exec --using 24.18.0 pnpm run knowledge:check -- --strict --format markdown`
- pre-push hooks

### DIFF
--- a/packages/core/src/mcp-advisor/tools/preview-api-endpoints.ts
+++ b/packages/core/src/mcp-advisor/tools/preview-api-endpoints.ts
@@ -5,6 +5,19 @@
 import { ObjectRegistry } from '../../registry.js';
 import type { ApiEndpoint, PreviewApiEndpointsInput } from '../types.js';
 
+const DEFAULT_EXAMPLE_BASE_URL = 'https://api.example.com';
+const EXAMPLE_UUID = '00000000-0000-4000-8000-000000000001';
+
+interface EndpointPreviewField {
+  name: string;
+  type: string;
+  required: boolean;
+  description?: string;
+  example: unknown;
+  readonly: boolean;
+  sensitive: boolean;
+}
+
 /**
  * Preview API endpoints that would be generated for a class
  */
@@ -12,7 +25,11 @@ export async function previewApiEndpoints(
   input: PreviewApiEndpointsInput,
 ): Promise<{ endpoints: ApiEndpoint[]; basePath: string; markdown: string }> {
   try {
-    const { className, basePath = '/api/v1' } = input;
+    const {
+      className,
+      basePath = '/api/v1',
+      baseUrl = DEFAULT_EXAMPLE_BASE_URL,
+    } = input;
 
     // Get class configuration
     const config = ObjectRegistry.getConfig(className);
@@ -27,6 +44,7 @@ export async function previewApiEndpoints(
       typeof apiConfig === 'object' && apiConfig?.exclude
         ? apiConfig.exclude
         : [];
+    const writableAllowlist = getApiWritableAllowlist(apiConfig);
 
     const shouldInclude = (endpoint: string) => {
       if (included && !included.includes(endpoint)) return false;
@@ -47,8 +65,14 @@ export async function previewApiEndpoints(
             ? field._meta.description
             : undefined,
         example: sampleValueForType(type, name),
+        readonly: isReadonlyField(field),
+        sensitive: isSensitiveField(field),
       };
-    });
+    }) satisfies EndpointPreviewField[];
+    const filterableFields = fieldsList.filter((field) => !field.sensitive);
+    const writableFields = fieldsList.filter((field) =>
+      isWritableRequestField(field, writableAllowlist),
+    );
 
     // Generate endpoint definitions
     const endpoints: ApiEndpoint[] = [];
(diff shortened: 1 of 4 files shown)


packages/core/src/mcp-advisor/tools/preview-api-endpoints.ts
```diff
@@ -279,6 +299,15 @@ function sampleValueForType(type: string, name: string): unknown {
     return [];
   }
 
+  if (
+    normalizedType.includes('foreignkey') ||
+    normalizedType.includes('crosspackageref') ||
+    normalizedType.includes('uuid') ||
+    normalizedName === 'id'
```
willgriffin, line 306: This narrowed check (`normalizedName === 'id'`) is a regression vs. the old `endsWith('id')` heuristic for a real code pattern. Cross-package foreign keys are intentionally declared as **plain `string`** rather than `type: 'foreignKey'` to avoid circular deps — this is documented in `packages/core/src/decorators/index.ts:250-272`:

```ts
// Cross-package: use a plain string instead
authorId: string = ''; // plain string — no circular dep
```

A field like `authorId`/`customerId`/`orderId` declared as `string` (no `type`) now falls through to `example-author-id` instead of the UUID shape, which is the common cross-package case. Worth restoring a name-based heuristic alongside the type check — e.g. `normalizedName === 'id' || normalizedName.endsWith('id')` — or at least `endsWith('Id')`/`endsWith('_id')`. The fixture in this PR (`customerId` with `type: 'foreignKey'`) happens to mask it because the FK type branch catches it first.

packages/core/src/mcp-advisor/tools/preview-api-endpoints.ts
```diff
@@ -290,9 +319,9 @@ function toKebabCase(value: string): string {
     .toLowerCase();
 }
 
-function buildCurlExample(endpoint: ApiEndpoint): string {
+function buildCurlExample(endpoint: ApiEndpoint, baseUrl: string): string {
   const path = endpoint.path.replaceAll(':id', 'example-id');
-  const url = new URL(`http://localhost:3000${path}`);
+  const url = new URL(path, baseUrl);
```
willgriffin, line 324: Minor: `new URL(path, baseUrl)` throws a bare `TypeError [ERR_INVALID_URL]` if a caller passes a `baseUrl` without a scheme or with stray whitespace (the runtime already wraps this in a generic `Failed to preview API endpoints: ...` rethrow at line 253, so it won't crash a caller — just a noisy message). Since `baseUrl` is user input, trimming it and/or validating the scheme here would give a clearer error. Low priority given the outer try/catch.

packages/core/src/mcp-advisor/types.ts
```diff
@@ -129,6 +129,7 @@ export interface ValidationResult {
 export interface PreviewApiEndpointsInput {
   className: string;
   basePath?: string;
+  baseUrl?: string;
```
willgriffin, line 132: Confirming what Copilot/Codex flagged: `baseUrl` is accepted by the implementation but **not** advertised in the MCP tool registration at `packages/core/src/mcp-advisor/index.ts:295-312` — that schema still only lists `className` + `basePath`. Schema-driven clients won't discover `baseUrl`, and strict validators will reject it, so the default host is effectively the only reachable one over MCP. Adding it to the `inputSchema.properties` (with a description noting it must be an absolute URL incl. scheme) would close the loop.

willgriffin (commented):
Reviewed the policy mirroring against the runtime REST generator (`packages/core/src/generators/rest.ts:1016` `applyWritablePolicy`) and the collection-layer sensitive-filter guard (`packages/core/src/collection.ts:460`).

The writable/sensitive logic tracks the runtime well, and I like that the preview now mirrors `api.writable` + server-managed + readonly exclusions so generated curls are actually submittable. Two non-blocking issues worth addressing: (1) the narrowed `id` heuristic regresses cross-package FK fields declared as plain `string`, and (2) `baseUrl` isn't advertised in the MCP tool `inputSchema` so clients can't discover it (already flagged by Copilot/Codex — confirming it's a real gap).

I confirmed the new `mirrors sensitive filter and writable body policies` test passes; the one unrelated test failure in this file also fails on `main` because the test manifest isn't generated in isolation — not a regression here.

Thanks for the iteration on #1838's feedback.

PR #1850, opened 2026-07-05
[codex] guard ShellState mutations from Svelte tracking

### LINKED ISSUE #1849: Guard ShellState mutations against Svelte effect tracking
## Summary

`ShellState.registerFocusTool()` mutates `$state` while also reading shell state. When a downstream app calls it from a Svelte `$effect` to mirror route/resource-specific tools into `AdminShell`, those internal `$state` reads can become dependencies of the caller effect. The subsequent write retriggers the same effect and can produce `effect_update_depth_exceeded`.

## Repro shape

A consumer does roughly this:

```svelte
$effect(() => {
  const unregister = tools.map((tool) => shell.registerFocusTool(tool));
  return () => unregister.forEach((cleanup) => cleanup());
});
```

With `@happyvertical/smrt-svelte@0.38.4`, this triggered Svelte update-depth errors during the willgriffin.dev AdminShell migration. The downstream compatibility patch was to wrap the shell mutations in `untrack(...)` at the call site.

## Expected

Public imperative mutation APIs on `ShellState` should be safe to call from consumer effects without leaking internal `$state` reads into the consumer tracking scope.

Likely candidates:

- `registerFocusTool()`
- `unregisterFocusTool()`
- `openFocusTool()`
- possibly panel mutation helpers that read before writing

## Notes

There is existing precedent in `tools-dock/define-tools-dock.svelte.ts`: `emitChange()` wraps `$state` reads in `untrack`, with a regression test describing this same class of reactive-read leak. `ShellState` should get an equivalent guard and regression coverage so downstream apps do not need compatibility wrappers.

willgriffin:
Fixes #1849.

## Summary

- Wrap ShellState imperative mutation paths in Svelte `untrack(...)` so internal `$state` reads do not become dependencies of consumer `$effect` scopes.
- Cover focus-tool, panel, and activity mutation paths with regression tests that mirror a consumer effect calling ShellState APIs.

## Root Cause

`ShellState` mutation methods read `$state` while also writing it. When downstream apps called methods like `registerFocusTool()` from a Svelte `$effect`, those internal reads could be captured by the caller effect. Later shell writes then retriggered the caller effect and could lead to `effect_update_depth_exceeded` loops.

## Review Notes

- Kept read APIs such as `snapshot()`, `listActivities()`, and `activityBadge()` trackable for normal `$derived` usage.
- Guarded mutation helpers, including settings, panels, focus tools, and activity lifecycle methods.
- Added a small Svelte harness that executes a supplied ShellState mutation from a real `$effect` and asserts later shell mutations do not increase the effect run count.

## Validation

- `biome check packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts packages/smrt-svelte/src/components/workspace/__tests__/admin-shell-state.test.ts packages/smrt-svelte/src/components/workspace/__tests__/admin-shell-mutation-effect-harness.svelte`
- `pnpm --dir packages/smrt-svelte exec vitest run src/components/workspace/__tests__/admin-shell-state.test.ts` (11 tests)
- `pnpm --filter @happyvertical/smrt-svelte typecheck`
- `pnpm --filter @happyvertical/smrt-svelte test` (48 files, 488 tests)
- `pnpm --filter @happyvertical/smrt-svelte build`
- pre-commit hooks: format, knowledge-check, secret-scan
- pre-push hooks: package DAG, format-check, knowledge-check, token/style guards, workflow validation

### DIFF
--- a/packages/smrt-svelte/src/components/workspace/__tests__/admin-shell-mutation-effect-harness.svelte
+++ b/packages/smrt-svelte/src/components/workspace/__tests__/admin-shell-mutation-effect-harness.svelte
@@ -1,0 +1,52 @@
+<script lang="ts">
+/**
+ * Runs a caller-supplied ShellState mutation inside a consumer $effect. Tests
+ * use this to prove ShellState's internal reads do not become dependencies of
+ * the consumer effect.
+ */
+import type { ShellState } from '../admin-shell/state.svelte.js';
+import type { ShellActivity } from '../admin-shell/types.js';
+
+type EffectCleanup = () => void;
+
+interface MutationInputs {
+  activity: ShellActivity;
+}
+
+interface Props {
+  shell: ShellState;
+  run: (shell: ShellState, inputs: MutationInputs) => EffectCleanup | undefined;
+  onReady: (controls: {
+    getRuns: () => number;
+    setActivityProgress: (progress: number) => void;
+  }) => void;
+}
+
+const props: Props = $props();
+const activity = $state<ShellActivity>({
+  id: 'reactive-sync',
+  label: 'Reactive sync',
+  kind: 'sync',
+  scope: 'system',
+  status: 'running',
+  progress: 0,
+});
+
+// Plain let on purpose: making this $state would add tracking noise to the
+// effect-leak signal this harness exists to measure.
+let runs = 0;
+
+$effect(() => {
+  runs += 1;
+  return props.run(props.shell, { activity });
+});
+
+// This harness intentionally snapshots the callback during setup.
+// svelte-ignore state_referenced_locally
+props.onReady({
+  getRuns: () => runs,
+  setActivityProgress: (progress: number) => {
+    activity.progress = progress;
+  },
+});
+</script>
--- a/packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
+++ b/packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
@@ -1,3 +1,4 @@
+import { untrack } from 'svelte';
 import {
   LocalStorageShellSettingsAdapter,
   mergeShellSettingsDelta,
@@ -24,6 +25,7 @@ import type {
 import { PANEL_EDGES, SCOPE_EDGES } from './types.js';
 
 type ActivityListener = (event: ShellActivityEvent) => void;
+type ShellActivityPatch = Partial<Omit<ShellActivity, 'id'>>;
 
 export interface ShellStateOptions {
   config?: ShellPanelDefaults;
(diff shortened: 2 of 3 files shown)


packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
```diff
@@ -107,24 +110,28 @@ export class ShellState {
...
   setPanel(edge: PanelEdge, state: VisiblePanelState): void {
-    if (this.config.panels[edge].initial === 'hidden') {
-      this.panels[edge] = 'hidden';
-      return;
-    }
-    if (state === 'expanded') this.closeExclusivePeers(edge);
-    this.panels[edge] = state;
-    this.settings = mergeShellSettingsDelta(this.settings, {
-      panels: { [edge]: state },
+    untrack(() => {
```
willgriffin, line 118: Non-blocking — maintainability note.

The thin delegators `expandPanel`, `collapsePanel`, `setHotkeysEnabled`, and `setHotkey` are safe today because they have no own `\` reads and funnel through this wrapped `setPanel`/`applySettings`. But that safety is implicit and uncovered by the new effect-leak tests.

If someone later adds a guard like `if (this.panels[edge] === 'hidden') return;` directly inside one of those delegators (a natural-looking change), it would reintroduce the leak with no test catching it. Two cheap options if you want belt-and-suspenders:

- a one-line `// safety: relies on setPanel()'s untrack — do not read \ here` comment on the delegators, or
- wrap the delegator body anyway (`untrack(() => this.setPanel(edge, 'expanded'))`) for defense-in-depth.

Not required for this PR.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
```diff
@@ -81,17 +82,19 @@ export class ShellState {
...
-      this.panels[edge] = resolveInitialPanelState(
-        edge,
-        this.config.panels[edge],
-        this.settings,
-      );
-    }
-    this.activeFocusToolId =
-      this.settings.activeFocusToolId ?? this.activeFocusToolId;
-    if (options.persist !== false) void this.persistSettings();
+    untrack(() => {
```
willgriffin, line 88: Non-blocking — test-coverage gap.

`applySettings` is the most read-heavy mutation (loops `PANEL_EDGES`, reads `this.settings` repeatedly, calls `resolveInitialPanelState`), so it's the highest-risk path for a future tracking leak. Its `untrack` wrapper is currently only covered transitively (the panel/focus-tool/activity tests exercise other methods).

Consider adding a fourth harness test of the shape:

```ts
const harness = mountMutationEffect(shell, (state) => {
  state.applySettings({ panels: { right: 'expanded' } });
  return undefined;
});
// ...later external applySettings / setHotkey shouldn't bump getRuns()
```

This would directly pin the delegation chain `setHotkey → applySettings` and catch a regression where someone unwraps `applySettings` thinking the callers cover it. Low priority — the current tests do prove the pattern works.

packages/smrt-svelte/src/components/workspace/__tests__/admin-shell-mutation-effect-harness.svelte
```diff
@@ -0,0 +1,30 @@
...
+type EffectCleanup = () => void;
+
+interface Props {
+  shell: ShellState;
+  run: (shell: ShellState) => EffectCleanup | undefined;
+  onReady: (controls: { getRuns: () => number }) => void;
+}
+
+const props: Props = $props();
+let runs = 0;
```
willgriffin, line 37: Nice harness design. Two observations, both non-blocking:

1. `runs` as a plain `let` (not `$state`) is the right call — making it `$state` would itself become an effect dependency and muddy the signal. Worth a one-line comment so a future reader doesn't 'helpfully' upgrade it: `// plain let on purpose — reactive runs would create its own tracking noise`.

2. The harness proves "no leak" by asserting `getRuns() === 1` after external mutations. That's the correct assertion, but it's worth noting it can't distinguish 'no tracking' from 'tracked but idempotent' — if a future ShellState mutation became a pure no-op on re-entry, this harness would pass even with a leak. The existing `expect(shell.panels.right).toBe('collapsed')` style assertions in each test cover the 'did the mutation actually happen' half, so the combination is sound. Just flagging the property-test boundary.

Good use of `state_referenced_locally` ignore for the init-time `props.onReady` read.

willgriffin (commented):
## Deep review

Verified end-to-end by checking out the PR, building the workspace, and running the suite against both the patched and unpatched source.

**Verdict: looks correct and well-scoped.** Approving with a few non-blocking notes.

### What I verified

- **The regression tests genuinely reproduce the bug.** Reverting `state.svelte.ts` to `main` while keeping the new tests produces the exact `effect_update_depth_exceeded` error from #1849 on all three new tests; restoring the patch makes them pass. The harness is not theatrical — it fails open.
- **Full suite green:** `vitest run` → 48 files / 488 tests pass; `tsc --noEmit` + `svelte-check` → 0 errors / 0 warnings.
- **Read APIs correctly stay reactive.** `snapshot()`, `listActivities()`, `activityBadge()`, and `homeEdgeForScope()` are intentionally not wrapped — confirmed they still track for `$derived` consumers. I added a throwaway `$derived` + `$effect` harness locally and confirmed `setPanel('right','expanded')` still notifies a `$derived(() => shell.panels.right)` reader (run count increments, value updates). So the `untrack` is surgical: it only blocks *internal* reads from leaking into a caller's effect scope; writes propagate normally.
- **Mutation-path coverage is complete.** Audited every public method that touches `$state`. Directly wrapped: `applySettings`, `setPanel`, `togglePanel`, `closeTopmostExpanded`, `registerFocusTool`, `unregisterFocusTool`, `openFocusTool`, `upsertActivity`, `updateActivity`, `removeActivity`. Delegating (no own `$state` reads): `expandPanel`/`collapsePanel` → `setPanel`; `setHotkeysEnabled`/`setHotkey` → `applySettings`; `hydrate` → `applySettings`. Helpers called from inside a wrapped scope (`closeExclusivePeers`, `homeEdgeForScope`, `emitActivity`, `persistSettings`) inherit the parent `untrack`. No gaps.
- **Matches the stated precedent** (`define-tools-dock.emitChange`'s `untrack` snapshot) in spirit; wrapping the whole method body is broader than the dock's targeted snapshot but appropriate here since ShellState mutations read+write more intertwinely.

### Non-blocking notes

Inline. Summary: (1) the delegation-only methods (`expandPanel`, `collapsePanel`, `setHotkeysEnabled`, `setHotkey`) are safe today but have no regression test guarding them against a future `$state` read sneaking in; (2) `applySettings` — the most read-heavy mutation — is covered transitively but not by a "called from `$effect`" test like the other three.

Neither needs to block merge.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
```diff
@@ -136,80 +143,96 @@ export class ShellState {
...
-        previous && previous.status !== normalized.status
-          ? 'transition'
-          : 'upsert',
-      activity: normalized,
-      previous,
+    untrack(() => {
+      const now = new Date().toISOString();
+      const normalized: ShellActivity = {
+        ...activity,
+        edge: activity.edge ?? this.homeEdgeForScope(activity.scope),
```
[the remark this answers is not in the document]
willgriffin: Fixed in `46805c7eb`: caller-supplied inputs are now snapshotted before entering `untrack`. `upsertActivity()` clones the external activity object before internal ShellState reads, `updateActivity()` snapshots patches, and the regression suite now proves a reactive activity `progress` change reruns the consumer effect and updates shell state.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
willgriffin: Addressed in `46805c7eb`: wrapped the thin delegators (`setHotkeysEnabled`, `setHotkey`, `expandPanel`, `collapsePanel`) in `untrack(...)` as defense-in-depth, so future local reads in those methods do not reintroduce the leak.

packages/smrt-svelte/src/components/workspace/admin-shell/state.svelte.ts
willgriffin: Addressed in `46805c7eb`: added a direct `applySettings()` effect-leak regression. The test calls `applySettings()` from the harness effect, then mutates through `setHotkey()` and verifies the effect run count stays stable while the keymap update still applies.

packages/smrt-svelte/src/components/workspace/__tests__/admin-shell-mutation-effect-harness.svelte
willgriffin: Addressed in `46805c7eb`: added the plain-`let` comment to the harness. The tests continue to pair run-count assertions with shell-state assertions, and the new reactive-activity-input test tightens the boundary further.

Issue #2382, opened 2026-08-17
Epic: database integration layer — close the schema-path / drift-oracle gap class (assessment 2026-08-17)


willgriffin:
# Epic: database integration layer — close the schema-path / drift-oracle gap class

> **Epic.** #2356 and #2358 turned out to be two instances of one structural condition: **the generator paths that produce production DDL are not the paths tests exercise, and nothing compares a live database with what the model layer assumes.** A 2026-08-17 assessment of core, tenancy, migrations, the DDL strategies, the runtime query path and the system tables found the same shape in ~25 places, several larger than the original two. This epic tracks all of them, plus #2356/#2357/#2358, with a parallel-agent gameplan and a two-release landing plan.
>
> Full report (ranked findings, `file:line` evidence, what checked out, method): https://claude.ai/code/artifact/b67f605d-19fe-45cb-bbc8-b840d39eb080 — a Markdown copy is the first comment on this issue so it survives in tracker history. The report is also input for agent tuning; the "what future agents should do differently" list lands in `core/AGENTS.md` via #2381 and in the `repo-happyvertical-smrt` Hindsight bank.

## Why nobody noticed

```
@smrt() class ─▶ scanner ─▶ manifest.json ─▶ generateSTISchemaFromManifest / generateCTISchemaFromManifest ─▶ registry.schema ─▶ db:migrate   ◀ PRODUCTION
                                                (no FK/xref indexes, STI unique dropped, foreignKeys [], triggers [])
ObjectRegistry fields ─▶ generateSchemaFromRegistry / generateSTISchemaFromRegistry ─▶ getTestDatabase() (195 test files)              ◀ TESTS
                                                (FK + STI partial FK indexes — the ones prod lacks)
AST ─▶ generateSchema ─▶ virtual smrt:schema (no consumers)                                                                            ◀ DEAD

Live-DB oracles: runtime = tableExists() only · doctor = no DB connection · db:status/db:diff = diff against the manifest above
⇒ an index the manifest never emitted is, by construction, "in sync".
```

Headline numbers (schemas regenerated from current source, 238 tables): **196/231 `@foreignKey` and 91/92 `@crossPackageRef` columns unindexed in production; 118 tenant-scoped tables whose slug is unique across tenants; 238/238 tables carry a redundant PK index; 0 DB foreign keys; 0 live-schema parity checks.**

## Scope / children

| # | Issue | Area | Sev | Size | Wave |
|---|---|---|---|---|---|
[...]

Issue #2363, opened 2026-08-17
core(schema): index the generated default list ordering — `(tenant_id, created_at)` for tenant-scoped tables, `created_at` otherwise


willgriffin:
## Problem

Generated REST lists default to `orderBy = 'created_at DESC'`, LIMIT 50 (`core/src/generators/rest.ts:1091`); fifteen collections across users/commerce/content/social hard-code the same. No schema path emits a `created_at` index (the dead AST path indexes `updated_at`). Combined with the missing tenant index every default list page on a tenant-scoped table is a full scan plus top-N sort — the query #2340 measured at 21 ms vs 0.1 ms with the right composite. Reports' incremental refresh scans `updated_at > watermark` with no index either.

## Change

Generator policy (all schema paths): tenant-scoped tables get `(tenant_id, created_at)`; others get `created_at`. Since PG scans B-trees backwards, no DESC declaration is needed. If the reference-column issue already emits a plain `tenant_id` index, prefer the composite and skip the standalone (a leading composite serves the tenant filter). Alternatively implement via #2357's declaration surface with a generator default — decide in the PR.

## Acceptance

- default REST list on a tenant-scoped table is an index scan with no sort (EXPLAIN attached from a PG snapshot)
- no duplicate standalone `tenant_id` index where the composite exists
- parity test updated

## Relationships

Same generator region as the reference-column issue — rebase after it; may use #2357. Deterministic ORDER BY / max limit for the generated routes are in the query-bounds issue.

From the 2026-08-17 database-layer assessment (finding A2).

---
**Part of epic #2382** (database-layer gap class; report in the epic's first comment). After #2359 (same generator region); may use #2357's declaration surface. Deterministic ORDER BY / max limit for generated routes are in #2367.

willgriffin:
```json
{"branch": "feat/issue-2363-default-list-ordering-index", "claimed_at": "2026-08-17T23:14:07.380535Z", "heartbeat_at": "2026-08-17T23:14:07.380562Z", "lease_seconds": 14400, "owner": "willgriffin", "runtime": "claude", "schema": "hv-agent-claim:v1", "session": "smrt-2363-list-ordering-index", "worktree": "/Users/will/Work/happyvertical/repos/smrt/.claude/worktrees/agent-a5220046e041b7ad1"}
```

willgriffin:
Implementation claim released by `claude` session `smrt-2363-list-ordering-index` (review).

willgriffin:
```json
[...]

PR #2394, opened 2026-08-17
fix(core): bound generated list queries and gate orderBy on protected fields

### LINKED ISSUE #2367: core: query bounds — `LIMIT 1` on get, default+max limit on list and generated routes, deterministic ORDER BY in the SvelteKit list route, orderBy sensitive-field check, chunked listByIds
## Problem

- `list()` has no default and no maximum limit; REST/SvelteKit default 50 with no cap (`core/src/generators/rest.ts:1089`, `sveltekit-generator.ts:2408`); `parseInt('abc')` → `LIMIT NaN` → 500. MCP caps in-process at 1000 but the runtime template does not (`mcp-runtime-template.ts:137`).
- `get()`/`findOne()`/SDK `db.get` emit no `LIMIT 1` on PG/SQLite (`collection.ts:1452`; `sql/dist/index.js:2481`); all matching rows travel through the driver to return `rows[0]`.
- The SvelteKit list route pages with LIMIT/OFFSET and **no ORDER BY** (`sveltekit-generator.ts:2397-2399`) → non-deterministic pages on PG (duplicates/skips); it also runs `count()` per page.
- `orderBy` is regex-validated (`collection.ts:1612`) but not checked against the field whitelist or the sensitive-field set, unlike `where` (#1540) and `select`: `?orderBy=api_secret&limit=1` is an ordering oracle.
- `listByIds()` (`collection.ts:1288`) does not chunk (PG 65535 / SQLite 32766 params) while include loaders and hierarchical chunk at 900.

## Change

- `LIMIT 1` on get/findOne (core and SDK); default limit and configurable max limit on `list()`; generated REST/SvelteKit/MCP-runtime clamp to the max and reject non-numeric.
- SvelteKit list route: default `ORDER BY created_at DESC, id` (deterministic tiebreak); consider optional keyset pagination later.
- `orderBy` validated against known, non-sensitive fields (same rail as `where`/`select`).
- Chunk `listByIds`.

## Acceptance

- generated list surfaces reject/clamp oversized limits and page deterministically
- get emits LIMIT 1 (SQL assertion)
- ordering by a sensitive field is rejected at the API boundary

## Relationships

Independent; first release wave. Touches collection.ts (different regions than the tenancy-read issue) and the generators.

From the 2026-08-17 database-layer assessment (finding E2).

---
**Part of epic #2382** (database-layer gap class; report in the epic's first comment). Independent; Release 1. Index for the default ordering is #2363. `collection.ts` region disjoint from #2365.

willgriffin:
Closes #2367

Part of epic #2382 (database-layer gap class), finding **E2**. Release 1.

## Problem

Four unbounded reads on the collection path, each reachable from a generated surface:

| | Before |
| --- | --- |
| `get()` / `findOne()` / `findById()` | `SELECT * FROM t WHERE …` with **no `LIMIT`**, then `rows[0]`. A filter not backed by a unique index streamed its whole match set through the driver and hydrated none of it. |
| `limit` / `offset` parsing | Three different hand-rolled parsers. `Number.parseInt('abc')` → `LIMIT NaN` (a **driver 500** for a caller typo). `Number(x) \|\| 50` folded a deliberate `limit=0` into 50. Nothing capped the value on REST or SvelteKit, so `?limit=100000000` was a full table scan plus a full hydration pass on a public endpoint. The MCP generator capped at 1000 in-process; the emitted stdio runtime did not cap at all. |
| `orderBy` | Checked only against `/^[a-zA-Z0-9_]+$/`. Not against the field whitelist, and **not against the sensitive-field set** that `where` (#1540) and `select` (#1902) both enforce. Ordering is a comparison, so `?orderBy=api_secret&limit=1` was an oracle over a column the same request may neither filter on nor project. |
| SvelteKit + MCP list surfaces | Paged with `LIMIT`/`OFFSET` and **no `ORDER BY` at all**. That is not pagination: PostgreSQL may return rows in any order and is free to choose a different one per query, so page 2 can repeat and skip page 1's rows — silently, and only under concurrency or a plan change. The emitted stdio MCP server was worse still: it built `collection.list({ where, limit, offset })` and never read `args.orderBy`, even though the generated tool schema advertises it — a caller passing `orderBy` got it **silently ignored**. |
| `listByIds()` | Expanded a caller-sized array into one `IN (…)` list with no chunking, past `SQLITE_MAX_VARIABLE_NUMBER` / PostgreSQL's 65535, failing the query before it ran. Every other IN-list loader (relationship, junction, hierarchy) already chunked at 900. |

## Change

`packages/core/src/query-bounds.ts` is the single bounds parser every surface now shares.
A bound is a non-negative integer or a **400-typed `QueryBoundsError`** (a `ValidationError`,
so `withRetry` never retries it and the existing `normalizeTypedHttpError` rail renders it as
a structured client error). Oversized pages **clamp** to `MAX_LIST_LIMIT` (1000) rather than
erroring, so a caller paging with too large a window still makes progress. An explicit `0`
means zero rows and is not folded into `DEFAULT_LIST_LIMIT` (50).

- **`get()` emits `LIMIT 1`.** Literal, not a bound parameter — it is not caller-derived and
  `$n` placeholders here would have to be numbered after the WHERE values on PostgreSQL.
- **`list()` validates `limit`/`offset`** and honours two new opt-in collection options,
  `defaultListLimit` and `maxListLimit`, validated at construction.
  There is deliberately **no implicit framework-wide default**: `list()` is the framework's
  bulk-read primitive and the relationship, junction, hierarchy and `listByIds()` callers all
  expect every matching row, so a default there would silently truncate correct queries rather
  than bound an attack surface. The bound belongs where untrusted input arrives, which is why
  the generated surfaces apply it unconditionally and programmatic callers opt in.
- **`orderBy` runs the same rail as `where` and `select`.** The whitelist/sensitive-set
  construction was extracted out of `convertWhereKeys()` into `collectQueryableFieldNames()`
  so the two cannot drift. Refused: `@field({ sensitive: true })`, `@field({ readPermission })`
  (redacted per caller by `toPublicJSON()` and already refused by `select`, so ordering over
  one is the same oracle through a different door), and fields that are registered but not
  column-backed — `oneToMany`/`manyToMany`/`meta`/`transient`, which pass a name whitelist but
  reach the driver as `no such column` and therefore surfaced as a **500** on a public
  endpoint. Protected columns are refused *before* the existence check, so the rejection reason
  for a secret never confirms the column exists. The client-facing message omits the
  valid-field list; the full list stays in `Error.message` for logs. `where`'s own behaviour is
  unchanged — the read-permission set is separate and consulted only by `orderBy`.
- **Every generated list surface orders deterministically**: `created_at DESC, <pk> ASC` on
  REST, SvelteKit, the in-process MCP generator and the emitted stdio MCP runtime.
  `created_at` alone still ties for rows created in the same tick, which is exactly what makes
  `LIMIT`/`OFFSET` paging repeat and skip. The tiebreak follows a declared
  `@field({ primaryKey: true })` (read from `_meta.primaryKey`, which is where scanner
  manifests carry it) because custom-primary-key classes have no synthetic `id` column at all.
  The stdio runtime cannot resolve a primary key on its own, so its per-object ordering is
  baked in at generation time via the new `RuntimeOptions.listOrderBy` map — the same shape
  `stiTargets` already uses. The matching index is **#2363**. An explicit `orderBy` still wins,
  and the stdio runtime now honours it at all.
- **`listByIds()` chunks** at `IN_LIST_CHUNK_SIZE` (900), and each chunk carries an explicit
  `limit` so an opt-in `defaultListLimit` cannot silently drop ids the caller asked for.
- `SmrtCollection.create()` widened from `SmrtClassOptions` to `SmrtCollectionOptions` so the
  new options are usable through the documented factory without a cast.

Keyset pagination is out of scope, as the issue specifies.

## Acceptance

| Issue acceptance criterion | Evidence |
| --- | --- |
| get emits `LIMIT 1` (SQL assertion) | `issue-2367-query-bounds.test.ts` spies `db.query` and asserts `/LIMIT 1\s*$/` on every emitted `SELECT` |
| generated list surfaces reject/clamp oversized limits | REST spec (400 on `abc`/`1.5`/`-1`/`1e3`, clamp to 1000), SvelteKit runtime test (executes the generated route: 400 before the collection is queried, clamp to 1000), MCP spec (tool error, clamp) |
| …and page deterministically | SvelteKit runtime test asserts `orderBy: ['created_at DESC','id ASC']` reaches `collection.list`; REST and MCP specs assert the same, plus the custom-primary-key variant on all three, plus the emitted stdio source |
| ordering by a sensitive field is rejected at the API boundary | REST spec: `?orderBy=apiSecret DESC` and `?orderBy=api_secret ASC` → 400 with a `sensitive` message; core test covers camelCase, snake_case, multi-term arrays, `readPermission` columns, and non-column-backed fields |

## Validation

| Check | Result |
| --- | --- |
| `packages/core` full `pnpm test` | **3407 passed**, 50 skipped, 0 failed (278 files, post-merge with main) |
| `packages/core` `tsc --noEmit` | 0 errors |
| `biome check` on every touched file | clean |
| `node scripts/check-coverage.mjs` | core (T1) **86.47%**, floor 80% — pass |
| `pnpm smrt dev:knowledge-check` | fresh, 0 errors |
| `node scripts/check-standards.mjs` | 64 packages, 0 violations |
| `node scripts/check-agents-chain.mjs` | 67 chains under cap |
| Downstream node-only suites (core consumers) | sales 427✓, commerce 229✓, tenancy 135✓, jobs 124✓, marketing 15✓ |

`packages/core` was rebuilt after the test runs so the `.smrt` knowledge artifacts are not stale.

**Not run locally, deferred to CI:** the PostgreSQL lane (no CI Postgres service in this
worktree; the change touches no schema, type, upsert target, timestamp or migration, so
SQLite masking is not a risk here), and the `users`/`agents`/`content`/`profiles` suites plus
several package builds, which fail in this agent worktree on a **pre-existing** environment
artifact — rolldown resolving through the parent checkout
(`Tsconfig not found …/repos/smrt/packages/accounts`, and `Could not resolve 'node:module'`
for the Svelte projects). Both reproduce unchanged on the untouched base commit
`6e310f27`.

## Review

Four independent review passes (two external reviewers plus the PR's automated reviewer) raised
nine findings. All were accepted and fixed, each with a regression test:

1. the REST tiebreak assumed an `id` column, breaking custom-primary-key models;
2. the SvelteKit check for a custom primary key read the top-level flag, but scanner manifests
   — its actual input — carry it under `_meta`, so it never matched;
3. `create()`'s parameter type (`SmrtClassOptions`) made the two new collection options
   unusable from type-checked consumer code, hidden internally by a cast because package
   typechecking excludes tests;
4. the MCP surfaces were left out of the ordering half entirely, and the emitted stdio runtime
   dropped the `orderBy` its own tool schema advertises;
5. `orderBy` accepted non-column-backed fields → a caller-triggerable 500;
6. `orderBy` refused `sensitive` but not `readPermission` columns — the same oracle;
7. REST's tiebreak resolver walked `getFieldMapsForPublicPolicy()`, which unions STI
   descendants, so a sibling's primary key could become this object's tiebreak — the query
   still succeeds on the shared STI table while ordering on a mostly-NULL column;
8. `id`/`slug`/`context` are whitelisted unconditionally but omitted from a custom-primary-key
   class's schema, so `?orderBy=id` on such a model was another caller-triggerable 500;
9. the whitelist is built with `toSnakeCase()` (which strips a leading underscore) while terms
   were normalised with `toDbColumnName()` (which keeps it). `SchemaGenerator` uses the
   stripped form, so a declared `_rank` field lives in a column called `rank`; `orderBy` now
   resolves each term to the column the schema actually names.

Residual risk noted and accepted: the emitted stdio MCP server is proven by asserting on the
exact generated lines rather than by executing it (unlike the SvelteKit route, which the tests
bundle and run). Standing up a runnable stdio-server harness is out of proportion to this
change.

```json
{"schema":"hv-agent-run:v1","policy_revision":"1.0.0","runtime":"claude-code","session":"smrt-2367-query-bounds-fix3","issue":"https://github.com/happyvertical/smrt/issues/2367"}
```


willgriffin:
Good catch — the two normalizations do disagree, and `orderBy` was the surface that made it reachable. Fixed in 8fe1f2531.

Checking which side is canonical settled it: `SchemaGenerator.toSnakeCase()` (`packages/core/src/schema/generator.ts:1620`) also **strips** the leading underscore, and it is what names the column (`:701`). So a declared `_rank` field lives in a column called `rank` — the whitelist's `toSnakeCase` form was right, and `toDbColumnName()`'s underscore-preserving form is correct only for the STI framework columns `_meta_type`/`_meta_data`, which really do carry the underscore and are never declared fields.

So rather than widening the whitelist (which would have let a term pass validation and then fail at the driver — the second half of your comment), `buildOrderBySql()` now resolves each term against the declared-column map first and falls back to the underscore-preserving form, and **emits the resolved name**:

```ts
const strippedName = toSnakeCase(field);
const columnName = definitionByColumn.has(strippedName)
  ? strippedName
  : this.toDbColumnName(field);
```

`orderBy: '_rank'` and `orderBy: 'rank'` now both emit `ORDER BY rank`; `_metaType` still reaches `_meta_type`. Covered by a new case in `issue-2367-query-bounds.test.ts` that declares a `_rank` field and asserts the emitted SQL for both spellings.

On line 513 (the STI ancestor loop) — same set, so it is fixed by the same reasoning: it keeps the `toSnakeCase` form, which is the column the generator emits.

On `convertWhereKeys()` (line 1124's rail): deliberately left alone. It emits the underscore-preserving form, so `where: { _rank: 1 }` is still rejected by the whitelist with a clear message rather than reaching the driver — the failure mode is a clean 4xx, not a 500. Changing what `where` emits would touch every internal hydration path (`loadFromId`, `getSavedId`, junction and relationship loaders all build `where` keys), which is a wider blast radius than this PR should carry. Noted in `packages/core/AGENTS.md` under Query bounds as a pre-existing gap of the same family.

